PR #26, opened 2017-02-09
Amp Start main page - Part 1

camelburrito:


css/www/index/page.css
```diff
@@ -16,4 +16,52 @@
 
 @import '../../core';
 /* Add components here */
-@import 'page-vars';
\ No newline at end of file
+@import 'page-vars';
+
+.www-header {
```
erwinmombay: why do we have the www prefix on these?

camelburrito: because these are www-* page specific styles.

erwinmombay: prefixing for framework/library styles makes sense but uncommon for page styles

PR #27, opened 2017-02-10
Main page  Part 2

camelburrito:
- Footer & Icons
- Navbar & Sidebar
- Phone preview


components/lightbox/template-preview.css
```diff
@@ -0,0 +1,51 @@
...
+ */
+@import '../../css/core-vars';
+
+.ampstart-iphone-mask {
+  background: var(--color-black);
+}
+
+.ampstart-iphone-frame {
+  border: solid 18px #e5ebee;
+  box-shadow: 0px 0px 0px 7px #bfc8ca;
```
erwinmombay: per style guide, should not have px if value is 0

components/lightbox/template-preview.css
```diff
@@ -0,0 +1,51 @@
...
+  margin: -315px 0 0 -180px;
+}
+
+.ampstart-iphone-frame::before,
+.ampstart-iphone-frame::after {
+  content: '';
+  display: block;
+  height: 50px;
+  width: 100%;
+  background: #e5ebee;
```
erwinmombay: see this repeated, why isnt it in a var somewhere?

camelburrito: This would not be a published component , it is used for creating a phone like enclosure around our templates being showcased, this does not change with a theme.

components/lightbox/template-preview.css
camelburrito: Done

components/lightbox/template-preview.css
erwinmombay: sure, but still best practice and all

PR #30, opened 2017-02-14
add highlight task

erwinmombay:


tasks/highlight.js
```diff
@@ -0,0 +1,36 @@
...
+const through = require('through2');
+
+function highlight() {
+  return gulp.src(config.src.www_code_pages)
+    .pipe(through.obj(function(file, enc, cb) {
+      if (file.isNull()) {
+        cb(null, file);
+        return;
+      }
+      file.contents = new Buffer(highlighter(file.contents.toString(),
```
camelburrito: did you explore if it was possible to replace the code in a div with a highlight? (if this is too hard we can work around it by using different source and dest folders :) )

tasks/config.js
```diff
@@ -27,6 +28,7 @@ module.exports = {
     default: 'dist',
     templates: 'dist',
     www_pages: 'dist',
+    www_code_pages: 'dist/www/code',
```
camelburrito: are we pushing it in the same file? we should put it in www/hl/<same-file-name> ?

tasks/highlight.js
```diff
@@ -0,0 +1,36 @@
...
+          'html'));
+      this.push(file);
+    }))
+    .pipe(gulp.dest(config.dest.www_code_pages));
```
camelburrito: how do we add the css file needed for the highlight ?

tasks/highlight.js
erwinmombay: not sure what you mean by code in a div, can you post a snippet?

tasks/config.js
erwinmombay: no idea, feel free to change after the fact or lmk what the path should be

tasks/highlight.js
erwinmombay: lets download https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.9.0/styles/default.min.css

camelburrito: @erwinmombay can you also put where you got this link from ?

erwinmombay: from https://highlightjs.org/download/

PR #76, opened 2017-02-22
Split JSON data into its own files.

camelburrito:


camelburrito:
CC @tonyruscoe @matharden @sebastianbenz @ericlindley-g 

This should make the lives of ppl working on templates or putting in actual content a lot better.

PR #86, opened 2017-02-24
bundle templates into archive with resources

erwinmombay:


camelburrito:
@erwinmombay - let me know when you need my review on it.

tasks/bundle.js
```diff
@@ -0,0 +1,64 @@
...
+const fs = require('fs-extra');
+const config = require('./config');
+
+
+function collectResources(filepath, html, done) {
+  const filename = path.basename(filepath, '.amp.html');
+  const env = jsdom.env(html, function(err, window) {
+    const ampimgs = window.document.querySelectorAll('amp-img[src]');
+    const imgs = [].slice.call(ampimgs).map(function(el) {
+      const src = el.getAttribute('src');
```
camelburrito: you will also need to parse srcset. I am okay with this for now. as we are not using srcset yet. But pls file a bug against yourself.

tasks/bundle.js
```diff
@@ -0,0 +1,64 @@
...
+const util = require('gulp-util');
+const exec = require('child_process').execSync;
+const jsdom = require('jsdom');
+const through = require('through2');
+const path = require('path');
+const fs = require('fs-extra');
+const config = require('./config');
+
+
+function collectResources(filepath, html, done) {
```
camelburrito: i dont see unminified CSS here. is that hard to do?

erwinmombay: should be easy. ill do it in follow up

tasks/bundle.js
erwinmombay: ack

PR #97, opened 2017-02-25
Support multi-line component anchors

sebastianbenz:
Use margin-bottom instead of height.

### DIFF
--- a/css/www/components/page.css
+++ b/css/www/components/page.css
@@ -35,7 +35,7 @@
 }
 
 .www-component-anchors > li {
-  height: var(--navbar-height);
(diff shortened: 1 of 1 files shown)


css/www/components/page.css
```diff
@@ -35,7 +35,7 @@
 }
 
 .www-component-anchors > li {
-  height: var(--navbar-height);
+  margin-bottom: var(--space-3);
 }
```
camelburrito: fixed it in a PR , made it use MB3 classes on the li's HTML - should be updated in master now

camelburrito:
Closing as this is obsolete now.

PR #98, opened 2017-02-25
Style anchors in component desc

sebastianbenz:
Not sure though if it makes sense to add these rules here if you don't need them for ampstart...

### DIFF
--- a/css/www/components/page.css
+++ b/css/www/components/page.css
@@ -61,6 +61,11 @@ body .www-component-anchors .www-component-anchor[option][selected] {
   margin-bottom: var(--space-6);
 }
 
+.www-component-desc a {
+  text-decoration: none;
+  color: var(--color-font-primary:);
+}
+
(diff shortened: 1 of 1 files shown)


sebastianbenz:
...the rule is too generic as well. Will match <button><a ....</a></button>.

css/www/components/page.css
```diff
@@ -61,6 +61,11 @@ body .www-component-anchors .www-component-anchor[option][selected] {
   margin-bottom: var(--space-6);
 }
 
+.www-component-desc a {
+  text-decoration: none;
+  color: var(--color-font-primary:);
```
camelburrito: i think this should not be generic. 

I made a similar change here 
https://github.com/ampproject/ampstart/blob/master/css/www/howitworks/page.css#L25

May be we could make a class i shared.css - www-anchor-highlight (or a bette name)? and reuse across?

PR #99, opened 2017-02-25
Disable pointer on disabled buttons

sebastianbenz:


sebastianbenz:
Changed `none`->`default`. This was actually a typo on my side and never intended. Sorry for the confusion. 

@derekcecillewis thanks for spotting this.

camelburrito:
Agreed it should be default, i guess that is what @derekcecillewis has been trying to say (I was thinking none was the default :))

camelburrito:
@sebastianbenz - i think your style should not be on :focus can you re do this?

sebastianbenz:
@camelburrito Why is that? I thought disabled elements can't accept focus.

PR #136, opened 2017-02-28
Add support for text-based logos in the navbar.

tonyruscoe:
Render the `amp-img` element only when the logo `url` has been provided, otherwise render an `h1` containing the site name instead. The default class names I’ve applied to the `h1` are what I need for the opinionated theme 2 template, but I could pass them in the `classes` variable instead.

Let me know what you think.

### DIFF
--- a/components/navbar/navbar.snip.html
+++ b/components/navbar/navbar.snip.html
@@ -28,7 +28,7 @@ limitations under the License.
         <amp-img src="{{url}}" width="{{width}}" height="{{height}}" layout="fixed" class="my0 mx-auto {{classes}}" alt="{{name}}"></amp-img>
       {{/url}}
       {{^url}}
-        <h1 class="h3 bold caps mx-auto {{classes}}">{{name}}</h1>
+        <div class="h3 bold caps mx-auto {{classes}}">{{name}}</div>
       {{/url}}
       {{#href}}
       </a>
--- a/components/navbar/navbar.snip.html
+++ b/components/navbar/navbar.snip.html
@@ -22,12 +22,17 @@ limitations under the License.
     <div role="button" on="tap:{{sidebar-id}}.toggle" tabindex="0" class="ampstart-navbar-trigger {{^sidebaronly}}md-hide lg-hide{{/sidebaronly}} pr2">☰</div>
     {{#logo}}
       {{#href}}
-      <a href="{{.}}" class="text-decoration-none inline-block">
+      <a href="{{.}}" class="text-decoration-none inline-block mx-auto">
       {{/href}}
-      <amp-img src="{{url}}" width="{{width}}" height="{{height}}" layout="fixed" class="my0 mx-auto {{classes}}" alt="{{name}}"></amp-img>
+      {{#url}}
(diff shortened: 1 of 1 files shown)


components/navbar/navbar.snip.html
```diff
@@ -22,12 +22,17 @@
     <a role="button" on="tap:{{sidebar-id}}.toggle"class="ampstart-navbar-trigger {{^sidebaronly}}md-hide lg-hide{{/sidebaronly}} pr2">☰</a>
     {{#logo}}
       {{#href}}
-      <a href="{{.}}" class="text-decoration-none inline-block">
+      <a href="{{.}}" class="text-decoration-none inline-block mx-auto">
```
camelburrito: why do we need`mx-auto` ?

components/navbar/navbar.snip.html
```diff
@@ -22,12 +22,17 @@
...
       {{/href}}
-      <amp-img src="{{url}}" width="{{width}}" height="{{height}}" layout="fixed" class="my0 mx-auto {{classes}}" alt="{{name}}"></amp-img>
+      {{#url}}
+        <amp-img src="{{url}}" width="{{width}}" height="{{height}}" layout="fixed" class="my0 mx-auto {{classes}}" alt="{{name}}"></amp-img>
+      {{/url}}
+      {{^url}}
+        <h1 class="h3 bold caps mx-auto {{classes}}">{{name}}</h1>
```
camelburrito: Lets change it to `<header>`, its probably not h1

tonyruscoe: It’s quite common for site names to be inside an `<h1>` but I don’t object to changing it to `<div>` or `<p>` if you prefer (it’s already inside a `<header>` element).

components/navbar/navbar.snip.html
tonyruscoe: To center the `<h1>` when the `href` is provided.

camelburrito: what happens if there are items on the `<nav>` that follows this (and the nav is being displayed.)

camelburrito: could you see what happens when there are navitems on the nav bar

components/navbar/navbar.snip.html
camelburrito: It can be h1 , its just extra effort to style it to not look like h1 :), you can change it to a div or span

components/navbar/navbar.snip.html
tonyruscoe: It doesn’t seem to make any difference.

components/navbar/navbar.snip.html
tonyruscoe: Changed to a `<div>`.

camelburrito:
LG please generate the snippets

tonyruscoe:
I’ve rebased and loaded the pages in www/render/ but there are no changes (apart from the indentation) because the test data used to generate those snippets doesn’t include an `href` for either variation of the navbar component.

I guess the components page would ideally show all possible variants of a component to ensure full coverage, but that’s not currently the case.

Given that, is this good to merge?

camelburrito:
I guess so ,  think it would be good to include it in the render/ - you can do it later in a separate PR

PR #137, opened 2017-02-28
Initial commit for Opinionated Templates/Theme 2.

tonyruscoe:
This PR includes the “home” and “menu” templates which should be mostly close enough to the mocks, but there are some known issues:

* sidebar uses default styles instead of how it was in the mocks
* “Fullpage Hero” component isn’t used since the markup wasn’t suitable
* `ampstart-dropcap` style isn’t used since it would have required completely overriding
* some minor spacing tweaks will probably be needed
* text logo is missing from the header navbar (requires #136 to be merged)

Let me know if you’d like any changes before merging.


css/templates/themes/2/page.css
```diff
@@ -0,0 +1,89 @@
...
+@import '../../../../components/images/fullpage-hero';
+@import '../../../../components/images/images';
+@import '../../../../components/navbar/navbar';
+@import '../../../../components/social-follow/social-follow';
+@import 'page-vars';
+
+body {
+  line-height: 1.75;
+}
+
```
camelburrito: may be consider rewriting this scale in your config so that you dont have to do this (look at `bass-override-vars`)

``` css
  --line-height-1: 1rem;
  --line-height-2: 1.125rem;
  --line-height-3: 1.5rem;
  --line-height-4: 2rem;
```

css/templates/themes/2/page.css
```diff
@@ -0,0 +1,89 @@
...
+h2,
+h3 {
+  font-family: 'Playfair Display', serif;
+}
+
+.h1,
+h1 {
+  font-size: 2.25rem;  /* 36/16 */
+  letter-spacing: .1px;
+  line-height: 1.33333;  /* 48/36 */
```
camelburrito: look into headings.css, 

again you could just override variables and do this (i guess except for letter spacing.). Same below

css/templates/themes/2/page.css
```diff
@@ -0,0 +1,89 @@
...
+.h6,
+h4,
+h5,
+h6 {
+  font-family: 'Lato', sans-serif;
+  letter-spacing: 0;
+  line-height: 1.25;  /* 20/16 */
+}
+
+.divider::after {
```
camelburrito: namespace - like 

`themename-divider`

css/templates/themes/2/page.css
```diff
@@ -0,0 +1,89 @@
...
+  border-bottom: 1px solid var(--color-border);
+  content: '';
+  display: block;
+  height: 2rem;
+  width: 8.33%;
+}
+
+article > h2[id] {
```
camelburrito: this could affect other stuff like accodions etc , lets namespace it , feel free to create a page level namespace to avoid collision. 

so you could do
```css
.namespace-section-class article > h2[id] {
}
```

css/templates/themes/2/page.css
```diff
@@ -0,0 +1,89 @@
...
+  /* Offset position of anchored headings to compensate for the header bar. */
+  margin-top: -5.5rem;
+  padding-top: 5.5rem;
+}
+
+article > h2 + p::first-letter {
```
camelburrito: is this drop cap , in that case just override the drop cap class here
ampstart-dropcap.

camelburrito:
I deally you should try and reuse the components so that if some one chooses your template and copy pastes code from `www/components.html` it should work

css/templates/themes/2/page.css
tonyruscoe: I’ve tried this locally and it’s caused quite a few problems, because of the way these variables are being (ab)used in other places, such as the `--navbar-height` variable and the `line-height` in the `.ampstart-navbar-trigger` ruleset.

Just because an element has the same height as another one doesn’t necessarily mean they should be using the same variable, unless there’s a design requirement for the dimensions to be explicitly dependent on one another.

In this case, I don’t think the `--navbar-height` variable and `.ampstart-navbar-trigger` styling should be dependent on either the `--line-height-6` or `--h2` variables, since those should be used exclusively for header styling.

How do you this should be resolved?

css/templates/themes/2/page.css
tonyruscoe: As explained above, this caused issues with the navbar.

That aside, how would you feel about turning `letter-spacing` into variables?

css/templates/themes/2/page.css
tonyruscoe: Given the CSS is inlined in the template and the namespace would be applied to the container, wouldn’t the risk of affecting other stuff be exactly the same? How would the namespace help here?

Adding a namespaced class to the `h2` sounds like it might address your concerns though. What do you reckon?

css/templates/themes/2/page.css
tonyruscoe: Yes, it’s for the drop cap. As explained in the PR description, I would have had to reset/override all of the declarations though. Explicitly including CSS which is going to get completely overridden feels like an odd recommendation. Is there an alternative? Perhaps creating a theme-namespaced version of the `ampstart-dropcap` class?

It would also have required the data and template code to be more complex, since I only want to add the dropcap to the first paragraph after each heading. I’m fine with doing that if necessary though.

tonyruscoe:
> I deally you should try and reuse the components so that if some one chooses your template and copy pastes code from `www/components.html` it should work

I really did try :) but what’s the plan when the current components aren’t suitable? You can see the difference in markup I used for the `fullpage-hero` replacement, for example. Should the component be more configurable? Should we create another variation of the component which gets styled appropriately by any theme, including baseline? Or is this fine for now, given these templates are meant to be opinionated?

css/templates/themes/2/page.css
camelburrito: I kind of anticipated this issue with 
`--navbar-height`
And hence i made it a variable too 

so you could really override both to whatever you want

css/templates/themes/2/page.css
camelburrito: The reason is - we can enforce a particular structure to your section , but we cannot enforce developers to not add anymore articles on the pages with more h2s to achieve a something different.

camelburrito:
> I really did try :) but what’s the plan when the current components aren’t suitable? You can see the difference in markup I used for the fullpage-hero replacement, for example. Should the component be more configurable? Should we create another variation of the component which gets styled appropriately by any theme, including baseline? Or is this fine for now, given these templates are meant to be opinionated?

We should keep the HTML the same and hack CSS to make it look like the theme. This is exactly why they are called opinionated.

css/templates/themes/2/page.css
tonyruscoe: Having to override a variable to keep its value the same when changing a seemingly unrelated variable is pretty convoluted and counterintuitive though. Would you not consider changing the base `--navbar-height` variable so it’s a fixed value which no longer depends on the header variables instead?

css/templates/themes/2/page.css
tonyruscoe: But your suggestion is to add the namespace to the `main` element, which means any other `article` child elements would still get this treatment, which doesn’t really address your concerns.

I’m going to add the namespaced class to the `h2` instead, unless you can think of any reason not to do it that way?

css/templates/themes/2/page.css
tonyruscoe: Done. PTAL.

css/templates/themes/2/page.css
tonyruscoe: Added a class to the `h2` instead. PTAL.

css/templates/themes/2/page.css
tonyruscoe: Done. PTAL.

css/templates/themes/2/page.css
tonyruscoe: Overriding `--navbar-height` for now. PTAL.

css/templates/themes/2/page.css
tonyruscoe: I’m overriding `--h*` and `--line-height-*` variables now. PTAL.

tonyruscoe:
> We should keep the HTML the same and hack CSS to make it look like the theme. This is exactly why they are called opinionated.

If you want to keep the same HTML (which isn’t feasible in this instance, because the semantic structure needs to change) are you suggesting we should override the styles declared by the BaseCSS classes? If so, I really think that’s the wrong thing to do.

We can chat about this outside of this thread, but I think if AMP Start is using BaseCSS as its framework, those classes should _always_ be respected and _never_ overridden by `ampstart` styles. This would basically mean making sure your `ampstart` declarations always have lower specificity and appear before BaseCSS declarations in the compiled CSS.

Anyway, we should discuss all of this. Would it be possible to get this merged in the meantime and then iterate on it?

templates/themes/2/home.amp.html
```diff
@@ -0,0 +1,77 @@
...
+      <header class="absolute top-0 right-0 left-0 center">
+        {{{heading-html}}}
+      </header>
+      {{#cta}}
+        <div class="absolute right-0 bottom-0 left-0 center">
+          <a class="h5 px4 py3 m3 inline-block text-decoration-none border" href="{{url}}">{{text}}</a>
+        </div>
+      {{/cta}}
+    </figure>
+  {{/fullpage-hero}}
```
camelburrito: I think you could still use the existing fullpage hero component for this, 
(you could either pass in a class to the top figure -- to center ) or wrap this with a div and add center to that.
And add a CTA section in the mustache to render the button (Use anchor inside a button to get the style for free.)

camelburrito: Rest of it looks good to me.

camelburrito:
Hey Tony, I Just had one comment, Rest of your PR looks good , let me know if you can't get to it, ill merge this and iterate on this

PR #150, opened 2017-03-01
Blog template text updates

ericlindley-g:


ericlindley-g:
I think because the content is being tested here: https://github.com/ampproject/ampstart/blob/6dc33a89c1a74045a30808eb5cc5ae6c0679ee60/templates/test-all.amp.json  ?

camelburrito:
The JSON is invalid , someplace you broke the JSON.

templates/blog-recipe.amp.json
```diff
@@ -139,7 +113,6 @@
         {"facebook": "#"},
         {"instagram": "#"},
         {"pinterest": "#"},
```
camelburrito: remove this comma and that will fix it

ericlindley-g:
Removed —those commas get me every time...

PR #159, opened 2017-03-01
bundle css and license file into archive

erwinmombay:


css/core.css
```diff
@@ -14,9 +14,9 @@
  * limitations under the License.
  */
 
+/*! Bassplate | MIT License | http://github.com/basscss/bassplate */
```
camelburrito: this should not go before normalize

tasks/bundle.js
```diff
@@ -66,3 +86,52 @@ function bundle() {
 }
 
 gulp.task('bundle', bundle);
+
+const licenses = `
+# The MIT License (MIT)
+
```
camelburrito: Should we be adding a heading to associate license to the respective projects?

css/core.css
erwinmombay: we need both of them at the top

tasks/bundle.js
erwinmombay: not sure how it works honestly but it makes sense to do that

camelburrito: Could you just do 

Basscss | https://github.com/basscss/basscss/blob/master/LICENSE.md
---------------------------------
ACTUAL Licence Text

Normalize | https://github.com/necolas/normalize.css/blob/master/LICENSE.md
---------------------------------
Actual license text

PR #161, opened 2017-03-02
Fixes

camelburrito:
- [x] Fix accordion touch-ability
- [x] Remove user notification
- [x] Fix render frames for very small screens

### DIFF
--- a/css/www/components/page.css
+++ b/css/www/components/page.css
@@ -78,6 +78,7 @@ body .www-component-anchors .www-component-anchor[option][selected] {
   white-space: pre-line;
   background: var(--color-highlight-background);
   overflow-y: auto;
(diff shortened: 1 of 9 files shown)


css/www/components/page.css
```diff
@@ -103,3 +104,22 @@ body .www-component-anchors .www-component-anchor[option][selected] {
     margin-right: var(--space-2);
   }
 }
+
+/* Remove the spacing around content when the width is less than
+ * --min-body-width (315px) + space on left (24px) + space on right (24px).
+ */
+@media (max-width: 363PX) {
```
erwinmombay: small `px`

camelburrito: Done

PR #166, opened 2017-03-02
Rename “thescenic” files.

tonyruscoe:
For consistency with the other templates.


hl-partials/navbar-with-sidebar.html
```diff
@@ -1,7 +1,7 @@
 <!-- Start Navbar -->
 <header class="ampstart-headerbar fixed flex justify-start items-center top-0 left-0 right-0 pl2 pr4">
   <div role="button" on="tap:header-sidebar.toggle" tabindex="0" class="ampstart-navbar-trigger md-hide lg-hide pr2">☰</div>
-  <amp-img src="..&#x2F;img&#x2F;blog-thescenic&#x2F;header-logo.png" width="110" height="33" layout="fixed" class="my0 mx-auto " alt="The Blog"></amp-img>
+  <amp-img src="..&#x2F;img&#x2F;thescenic&#x2F;header-logo.png" width="110" height="33" layout="fixed" class="my0 mx-auto " alt="The Blog"></amp-img>
```
camelburrito: replace `&#x2F;` with `/`

camelburrito:
LG except for one nit

hl-partials/navbar-with-sidebar.html
tonyruscoe: This is a problem in many pages which I think is caused by using `{{var}}` instead of `{{{var}}}` for paths and URLs.

Since it’s not a problem introduced in this commit, I’ve raised #168 and will fix all templates as a follow-up.

tonyruscoe:
I’ve raised #168 for that nit, so this should be good to merge as-is.

PR #175, opened 2017-03-02
Open social follow in a new window.

camelburrito:


tonyruscoe:
I think this should be reverted. Some reasons:

- Adding `target=_blank` to social follow links stops the native mobile app intents from being triggered, at least on Chrome for Android (I’ve not tested other combinations of browser or OS).
- Opening links in new windows should be saved for scenarios where the user could lose data, e.g. sign-up flows, shopping carts, etc. The browser back button should work fine otherwise.

Related reading, some of which is old but still applicable:

https://www.nngroup.com/articles/the-top-ten-web-design-mistakes-of-1999/
https://www.nngroup.com/articles/top-10-mistakes-web-design/
https://www.smashingmagazine.com/2008/07/should-links-open-in-new-windows/

PR #177, opened 2017-03-02
text changes

spacedino:


templates/article.amp.html
```diff
@@ -24,7 +24,8 @@
   {{> ../components/images/fullpage-hero.snip.html}}
   <main id="content" role="main" class="">
     <article class="photo-article">
-      <h2 class="mb4 px3">Photo Gallery</h2>
+      <span class="ampstart-accent block px3 pt2 mb2">PHOTO GALLERY</span>
```
tonyruscoe: FWIW, this should semantically still be a heading and to make it upper case, you should add the `caps` class. If you’re not able to style it appropriately using the `.h1`–`.h6` classes, then those should probably be adjusted instead of making this a `span` element.

camelburrito: Per Abby this is not a heading - just checked with her - this is just a tag.

ericlindley-g: Maybe worth having a "category" or "tag" text style for blogs?

PR #192, opened 2017-03-03
Fonts for WWW

camelburrito:


css/ampstart-base/base-page.css
```diff
@@ -24,13 +24,19 @@ body {
...
   line-height: var(--line-height-3);
-  font-family: var(--font-family);
+  font-family: var(--base-font-family);
   min-width: var(--body-min-width);
   overflow-x: hidden;
   font-smooth: always;
   -webkit-font-smoothing: antialiased;
 }
 
+/* This is needed to override from templates*/
```
tonyruscoe: Out of interest, what are you trying to achieve here that wasn’t working with the previous method?

This `font-family` declaration will _always_ override the one in the ruleset above, even if it’s empty. If the font family declared here can’t be found, the default system font – likely serif – will be used instead, unless `--font-family` is a comma-separated list of alternatives (which it’s not in either www/page-vars.css or www/shared.css).

camelburrito: I was trying to have the fallback font be what we declared (Per UX). 
The build process kills this if there is no override and uses the body selector. Hence this works.

PR #194, opened 2017-03-03
Minor Font size and color fixes in WWW.

camelburrito:


css/www/shared.css
```diff
@@ -54,6 +59,7 @@ h1,h2,h3,h4,h5,h6 {
   bottom: 0;
 }
 
+
```
matharden: nit: no need for this extra line.

camelburrito: Fixed!

camelburrito:
Thanks for the review!

PR #195, opened 2017-03-03
Fix license + css file name

camelburrito:
Headings for licenses.

Write unminified css file as max.css (instead of amp.css) in the zip file

### DIFF
--- a/tasks/bundle.js
+++ b/tasks/bundle.js
@@ -88,6 +88,9 @@ function bundle() {
 gulp.task('bundle', bundle);
 
 const licenses = `
+Basscss | https://github.com/basscss/basscss/blob/master/LICENSE.md
+----------------------------------------------------------------------------
+
 # The MIT License (MIT)
 
 Copyright © Nicolas Gallagher and Jonathan Neal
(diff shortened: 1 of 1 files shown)


camelburrito:
@ericlindley-g - we might have to add some more wordings on `How it works` since we now have unminified CSS

ericlindley-g:
Gotcha — what's changed about what they need to do? Do they now always need to minify after modifying, or are we providing minified as well? And do the files all fit under the 50K limit anyway?

tasks/bundle.js
```diff
@@ -110,7 +113,8 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 
-
+Normalize | https://github.com/necolas/normalize.css/blob/master/LICENSE.md
```
tonyruscoe: Nit: I think the official project name is called “Normalize.css”

tasks/bundle.js
```diff
@@ -121,6 +125,8 @@ The above copyright notice and this permission notice shall be included in all c
 THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 
+AMP Start | https://github.com/ampproject/ampstart/blob/master/LICENSE
```
tonyruscoe: Is it “AMP Start” or “AMPStart” (as in the license below)? Should really be consistent.

PR #196, opened 2017-03-03
Reuse Fullpage hero on theme2 as much as possible

camelburrito:


camelburrito:
> but I do think we should think about a better way of making these components reusable without requiring quite so much markup and CSS overrides.

Yes - We will need to separate markup from bass- layout classes. I've been wanting to do it for some time.

PR #214, opened 2017-03-06
Updated "get started" language and remove non-loading Twitter

ericlindley-g:
/cc @spacedino


templates/article.amp.html
```diff
@@ -52,8 +52,6 @@
 
       <p class="mb4 px3">Integer fringilla eu lacus sit amet tristique. Nulla sagittis fermentum porta. Aenean convallis nulla eu nulla lobortis, sit amet sagittis mauris tincidunt. Praesent sed efficitur odio. Quisque et sollicitudin nisi. Fusce sed auctor ligula. Suspendisse potenti. Nam hendrerit sollicitudin enim, non aliquet lacus congue et.</p>
 
-      <amp-twitter width="390" height="330" layout="responsive" data-tweetid="819614235210620928" data-cards="hidden" class="ampstart-card mb4"></amp-twitter>
-
```
camelburrito: why are we getting rid of this?

ericlindley-g: Because it's not loading in the nested iframe for the template preview. Wouldn't be a problem in production, but looks broken until we can fix the amp-twitter bug

PR #234, opened 2017-03-07
Chevron - Quick fix

camelburrito:


ericlindley-g:
/cc'ing @spacedino  as FYI. Replaces "v" with unicode character for chevron. Will replace with proper image asset as soon as possible.

PR #242, opened 2017-03-08
Suggestion: make repo self-sufficient by using npm script aliases to the gulp tasks

kahlil:
This pull request aliases the gulp tasks mentioned in the readme to npm scripts calls. 
This makes the project independent from a global gulp install. 
Global installs as a dependency to a project can cause unnecessary problems when developers have to have incompatible versions globally installed or just don't want to install deps globally. 
Luckily nowadays it is absolutely not necessary to globally install any task runner because we have `npm scripts`.

I did not consult anybody before doing this, but since it is just a small change I thought I'll send it as a quick PR. Please just dismiss if you don't like this suggestion.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,13 +1,13 @@
 {
   "name": "kickstart",
-	"script": {
-		"build": "gulp build",
-		"clean": "gulp clean",
-		"www": "gulp www",
-		"highlight": "gulp highlight",
-		"watch": "gulp watch",
-		"serve": "gulp serve"
-	},
+  "script": {
+    "build": "gulp build",
+    "clean": "gulp clean",
+    "www": "gulp www",
+    "highlight": "gulp highlight",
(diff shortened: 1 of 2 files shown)


camelburrito:
@kahlil Thanks for doing this. I just came back from NYC after amp conf hence the delay! 

@erwinmombay can you take a look at this please.!

package.json
```diff
@@ -1,5 +1,13 @@
 {
   "name": "kickstart",
+	"script": {
+		"build": "gulp build",
```
erwinmombay: can you align to 2 indent? otherwise LGTM. thanks!

camelburrito:
also please update the readme file with changes needed.

kahlil:
will do

kahlil:
@camelburrito @erwinmombay fixed.

Issue #243, opened 2017-03-13
disallowed * (universal selector) in template css


timBrockman:
`*{box-sizing:border-box}` is being used on line 18 of page-base.css, line 1310 of article.max.css and blog.max.css (as well as in the amp-custom style of both templates).

What do you thing about removing this and explicitly adding it in the classes/elements that need it? It may add a little to the file size, but it is on the verboten list.

It also shows up in `.ampstart-image-heading>*{margin:0}` & `.ampstart-headerbar+amp-sidebar+*{margin-top:3.5rem}`

timBrockman:
I've done some investigating and I think the `*{box-sizing:borde-box}` could be replaced with:
```
amp-sidebar,
div,
section,
input,
textarea,
select,
label,
button,
.btn,
.border,
.border-top,
.border-right,
.border-bottom,
.border-left,
.ampstart-btn,
.ampstart-input {
box-sizing:border-box;
}
```
I ran a build and checked each template visually against `*{box-sizing:border-box}` and it looks ok...
I'm not sure if you want both the likely elements and classes, or if we want both ampstart classes and basscss classes.

I'll start on the other two *s.

timBrockman:
`.ampstart-image-heading > * {margin:0}` replaced with probable headings:

```
.ampstart-image-heading > .h1,
.ampstart-image-heading > .h2,
.ampstart-image-heading > .h3,
.ampstart-image-heading > .h4,
.ampstart-image-heading > .h5,
.ampstart-image-heading > .h6,
.ampstart-image-heading > .ampstart-subtitle,
.ampstart-image-heading > .caps,
.ampstart-image-heading > h1,
.ampstart-image-heading > h2,
.ampstart-image-heading > h3,
.ampstart-image-heading > h4,
.ampstart-image-heading > h5,
.ampstart-image-heading > h6 {
  margin: 0;
}
```
You may want fewer options. The Scenic example only uses h2.

timBrockman:
Replaced:
```
.ampstart-headerbar + *:not(amp-sidebar),
.ampstart-headerbar + amp-sidebar + * {
  margin-top: var(--navbar-height);
}
```
with:
```
.ampstart-headerbar + div,
.ampstart-headerbar + figure,
.ampstart-headerbar + section,
.ampstart-headerbar + article,
.ampstart-headerbar + main,
.ampstart-headerbar + nav,
.ampstart-headerbar + footer,
.ampstart-headerbar + header,
[...]

PR #291, opened 2017-03-14
replace disallowed universal selectors (#243)

timBrockman:
Replaced the 3 instances of the disallowed universal selector that were showing up in the starter templates with type &| class selectors likely to affected by those styles.
This could be improved with a little discussion.

### DIFF
--- a/components/images/images.css
+++ b/components/images/images.css
@@ -20,7 +20,20 @@
   background: linear-gradient(to top, rgba(0,0,0,0.65) 0%,rgba(0,0,0,0) 100%);
 }
 
-.ampstart-image-heading > * {
+.ampstart-image-heading > .h1,
+.ampstart-image-heading > .h2,
+.ampstart-image-heading > .h3,
+.ampstart-image-heading > .h4,
+.ampstart-image-heading > .h5,
+.ampstart-image-heading > .h6,
+.ampstart-image-heading > .ampstart-subtitle,
+.ampstart-image-heading > .caps,
+.ampstart-image-heading > h1,
+.ampstart-image-heading > h2,
+.ampstart-image-heading > h3,
+.ampstart-image-heading > h4,
+.ampstart-image-heading > h5,
+.ampstart-image-heading > h6 {
   margin: 0;
 }
 
--- a/components/navbar/navbar.css
+++ b/components/navbar/navbar.css
@@ -21,8 +21,22 @@
   box-shadow: 0 0 5px 2px rgba(0,0,0,.1);
 }
 
-.ampstart-headerbar + *:not(amp-sidebar),
-.ampstart-headerbar + amp-sidebar + * {
+.ampstart-headerbar + div,
+.ampstart-headerbar + figure,
+.ampstart-headerbar + section,
+.ampstart-headerbar + article,
+.ampstart-headerbar + main,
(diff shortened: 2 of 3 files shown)


components/images/images.css
```diff
@@ -20,7 +20,20 @@
...
+.ampstart-image-heading > .h6,
+.ampstart-image-heading > .ampstart-subtitle,
+.ampstart-image-heading > .caps,
+.ampstart-image-heading > h1,
+.ampstart-image-heading > h2,
+.ampstart-image-heading > h3,
+.ampstart-image-heading > h4,
+.ampstart-image-heading > h5,
+.ampstart-image-heading > h6 {
   margin: 0;
```
camelburrito: this could just go in the markup as bass classes

components/navbar/navbar.css
```diff
@@ -21,8 +21,22 @@
...
+.ampstart-headerbar + header,
+.ampstart-headerbar + amp-sidebar + div,
+.ampstart-headerbar + amp-sidebar + figure,
+.ampstart-headerbar + amp-sidebar + section,
+.ampstart-headerbar + amp-sidebar + article,
+.ampstart-headerbar + amp-sidebar + main,
+.ampstart-headerbar + amp-sidebar + nav,
+.ampstart-headerbar + amp-sidebar + footer,
+.ampstart-headerbar + amp-sidebar + header {
   margin-top: var(--navbar-height);
```
camelburrito: The usage here is intentional and is not bad since it is not global and it selects only one element.

css/ampstart-base/base-page.css
```diff
@@ -15,7 +15,22 @@
...
+button,
+.btn,
+.border,
+.border-top,
+.border-right,
+.border-bottom,
+.border-left,
+.ampstart-btn,
+.ampstart-input {
   box-sizing: border-box;
```
camelburrito: This is an intentional reset. Also remember we should almost never override bass  classes (especially not like this)

timBrockman: Agreed, especially about overriding bass classes. I got a bit confused about the specificity when I was playing with the dev tools. Could we get away with just reseting some likely to be bordered type selectors? People can add what isn't covered for their specific use.

components/navbar/navbar.css
timBrockman: Sounds good. I noticed other similar instances as well and the validator didn't seem to care. We should probably put it a PR to clarify the [amp docs](https://www.ampproject.org/docs/guides/author-develop/responsive/style_pages#disallowed-styles). I read the docs to be a blanket restriction on the use of `*` or `:not()` even though there were pretty specific reasons.

components/images/images.css
timBrockman: Good call.

camelburrito:
* selector is not that bad anymore . We  have to be conscious when using it.

http://www.telerik.com/blogs/css-tip-star-selector-not-that-bad

components/navbar/navbar.css
timBrockman: I think the original code is both an elegant & flexible solution to the margin-top with floating menu problem. I just didn't think it was allowed.

timBrockman:
It looks like the docs are going to be updated. Closing

Issue #338, opened 2017-03-15
The nav bar is eating the blue Food in the blog template.


timBrockman:
blog template [L27](https://github.com/ampproject/ampstart/blob/master/templates/blog.amp.html#L27)  is getting covered up by the navbar. We could remove `id="content"` and let the navbar styles take over or change the [`#content{margin-top:0, padding-top:var(--navbar-height)` ](https://github.com/ampproject/ampstart/blob/master/css/ampstart-base/base-page.css#L56).

timBrockman:
Looks to be resolved. Closing...

PR #341, opened 2017-03-15
removed id="content" (#338)

timBrockman:
This is one way to keep the navbar from covering the accent above the heading (L27). It looks like this id isn't being referenced by anything else. Another option would be changing the style, but both themes rely on `#content`.


timBrockman:
Good call. I noticed later that theme 2 menu followed a similar pattern as the blog template.

PR #359, opened 2017-03-22
UI components test page

spacedino:
added all UI components from copied and pasted code on our UI components page. This will be used for testing of any CSS to make sure the page is performing correctly.

NOTES for intended use:
- copied and pasted BELOW the head / minified CSS of a template and below template / theme content but still in body
- CSS of a template / theme can be applied to the head of this page BUT please do not update this file with theme css. This doc should remain as the base for the test, but not include specific css. We could have a theme test file saved with themes.

### DIFF
--- a/templates/test-all.amp.html
+++ b/templates/test-all.amp.html
@@ -66,6 +66,741 @@ limitations under the License.
 
   {{> ../components/form/form.snip.html}}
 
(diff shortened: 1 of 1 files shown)


ericlindley-g:
LGTM but deferring to @camelburrito for "real" review :)

camelburrito:
I modified this from your PR and fixed the errors and submitted it as another PR

PR #360, opened 2017-03-22
Make .ampstart-btn use body font

ithinkihaveacat:
The "VIEW TEMPLATE" and "DOWNLOAD CODE" buttons use a different font—"VIEW CODE" is Helvetica, whilst "DOWNLOAD CODE" is in Montserrat. (Interestingly this isn't true on https://ampstart.com, however the HTML is slightly different there—I'm guessing f7286b0d29db5289e227993fe39517345601d534 hasn't been deployed yet?)

This PR makes the font used by `<button class="ampstart-btn">…</button>` and `<a class="ampstart-btn">…</a>` match.

screen shot 2017-03-22 at 11 22 07 pm
screen shot 2017-03-22 at 11 22 18 pm

### DIFF
--- a/css/ampstart-base/base-page.css
+++ b/css/ampstart-base/base-page.css
@@ -41,6 +41,11 @@ p {
   margin: 0;
 }
 
+button {
+  font-family: inherit;
+  font-weight: inherit;
+}
+
 /* Subtitle */
 .ampstart-subtitle {
   color: var(--color-font-subtitle);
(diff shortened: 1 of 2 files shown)


css/ampstart-base/base-page.css
```diff
@@ -41,6 +41,11 @@ p {
   margin: 0;
 }
 
+button {
```
camelburrito: This should be changed in ampstar-btn classes in buttons.css

ithinkihaveacat: Moved in 1780b70.

Issue #354, opened 2017-03-17
Theme 2 - check boxes broken


spacedino:
needs check mark

![image](https://cloud.githubusercontent.com/assets/7339798/24062099/6690e1b2-0b17-11e7-8768-3b058592ba57.png)

spacedino:
[icons_check_menu_close.zip](https://github.com/ampproject/ampstart/files/851822/icons_check_menu_close.zip)

@ericlindley-g  to prioritize cc @camelburrito to fix

camelburrito:
# First Timers Only
We know figuring out the process for contributing to an open source project can be intimidating, so we created this issue as a way for you to learn the ropes. (If you feel comfortable contributing to open source projects, please leave this issue for someone else.)

This particular starter issue is designed to familiarize you with the AMP Start Code Contribution.

# Background
[www.ampstart.com](www.ampstart.com) is a website that helps web developers to be able to download templates that are valid AMPs and use them on their pages. 


# What to do
- [ ] Create a test page (you can use the [template page](https://github.com/ampproject/ampstart/blob/master/templates/themes/2/home.amp.html) and add a check box in it to test it out  
OR 
[test.amp.html](https://github.com/ampproject/ampstart/blob/master/templates/test-all.amp.html) and point the css to theme2's css in the [JSON Config](https://github.com/ampproject/ampstart/blob/master/templates/themes/2/home.amp.json#L14))

- [ ] Now that you have replicated the problem try and figure out why the tick mark is not rendered in the checkbox.

- [ ] Most likely reason is `--color-input-secondary` css variable is not set in the [page-vars.css](https://github.com/ampproject/ampstart/blob/master/css/templates/themes/2/page-vars.css#L53) for theme 2. Check out the variables that need to be overridden to solve this in `page-vars.css` in this[ component css var file](https://github.com/ampproject/ampstart/blob/master/components/inputs/input-vars.css#L21).

### Step by step
- [ ] Claim this issue by adding a comment below.  Please only claim this bug if you plan on starting work in the next day or so.
[...]

PR #439, opened 2017-05-17
Checkbox fix

Lillian-wang:
Fix the bug: Theme 2 - check boxes broken #354


css/templates/themes/2/page-vars.css
```diff
@@ -50,7 +50,7 @@
   --color-navbar-primary: var(--color-black);
   --color-navbar-font-primary: var(--color-white);
 
-  --color-input-primary: var(--color-font-accent);
+  --color-input-primary: var(--color-secondary);
```
camelburrito: Changing the primary color will break other inputs , can you try to set the right secondary color (i have mentioned the variables in the steps)

Lillian-wang: So I should keep " --color-input-primary: var(--color-font-accent);"  and add a new line: " --color-input-secondary: var(--color-secondary); " , right?

camelburrito: yes , it should be a color that fits the theme (i don't know what it is until i download your patch and check - if you can upload an img i can verify)

Lillian-wang: Ok updated. PTAL

camelburrito:
@Lillian-wang - do you have screenshots of how it looks after your fix?

Lillian-wang:
@camelburrito , can you specify what screenshot you would like to have? Since I am just adding some attributes onto links, the site looks exactly the same as before.

camelburrito:
@Lillian-wang - this is the checkbox bug you fixed not the links one :)

Lillian-wang:
@camelburrito , for some reason, it worked when I fixed the bug last week. But now it doesn't work again. I will take a look later.

Lillian-wang:
@camelburrito 
screen shot 2017-05-29 at 11 53 21 pm

camelburrito:
@Lillian-wang This looks great!  Thanks!

Issue #416, opened 2017-04-14
[A11Y] [Web] Incorrectly labeled Icon


shengoogle:
ENVIRONMENT:
Build number: 56.0.2924.87 beta (64-bit)
Form factor: Chromebook
Browser: Chrome

PRE-REQUISITE
chromevox is ON

STEPS TO REPRODUCE:
1. Open https://ampstart.com
2. Navigate to Components Tab
3. Using chromevox, navigate to Follow Icon
4. Cycle through Icons and select Google+ Icon

ACTUAL RESULTS:
Google+ icon reads as shape

EXPECTED RESULTS:
Icon should properly read Google+

ATTACHMENT:
Please see the video in the link below for reference.

https://drive.google.com/open?id=0BwUNSMCBrCB4TVpTUnVIVnJmQlU

comolongo:
The Google drive videos across all of the A11Y bugs require permission to view. Since this is a public project, can those videos be made public?

shengoogle:
I'm not sure, I'll have to check with my manager but I do not mind setting
it to be made public if it turns out it is ok.

Lillian-wang:
I'm working on this. I will have something out later this week.

torch2424:
It appears that @Lillian-wang fixed this issue, and after testing this myself, this is now correctly working in latests PROD, this can be closed

PR #446, opened 2017-05-31
Fixed the [A11Y] [Web] Incorrectly labeled Icon #416

Lillian-wang:
#416 
Renamed title tags in the svg icons from default 'Shape' to more specific descriptions, so chromevox can read them correctly.


Lillian-wang:
@camelburrito   It still builds fine on my machine. Not sure what's wrong.

PR #447, opened 2017-05-31
Fix for #420: Changed the min-width of sidebar to avoid text colliding with the edge of the side bar.

### LINKED ISSUE #420: remove side nav before 833 px wide
switch to tablet view sooner to avoid -OR- don't shrink sidebar to avoid the text in the sidebar colliding with the edge of the side bar

image

Lillian-wang:
#420 remove side nav before 833 px wide.
Changed the min-width of sidebar to avoid text colliding with the edge of the side bar.


Lillian-wang:
@camelburrito , is there some way to kick off the test again? I'm not sure where it broke. It builds fine on my machine. https://pastebin.com/NSergd0T

PR #453, opened 2017-06-06
Link menu in restaurant theme

robinpokorny:
The restaurant theme has two pages. The second one with the menu is undiscoverable.

This adds the links between these two pages.

### DIFF
--- a/templates/themes/2/home.amp.json
+++ b/templates/themes/2/home.amp.json
@@ -21,7 +21,7 @@
           "link": {"url": "#", "text": "Home"}
         },
         {
-          "link": {"url": "#", "text": "Menu"}
+          "link": {"url": "menu.amp.html", "text": "Menu"}
         },
         {
           "link": {"url": "#locations", "text": "Locations & Hours"}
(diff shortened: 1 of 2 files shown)


ericlindley-g:
For now, the "download" button at this time only downloads the first restaurant page, so this can create a situation where the user is looking at page 2, but downloads page 1 when hitting "download". 

May be worth dealing with this issue in the short term, to enable the user to see more of the theme at large— + @spacedino for UX feedback in light of the request

ericlindley-g:
By the way, thanks for the contribution, @robinpokorny !

robinpokorny:
@ericlindley-g, that is true. The correct way would IMO be that the zip includes both pages. A brief look at the code tells me it would be a bigger change.

I'm glad I can contribute and hope to add more on the future!

ericlindley-g:
Agreed :)  I should have added that we're hoping to include all files from the "site" in a zip, but we've been working through other items on the priority list first

/cc @erwinmombay for visibility

Issue #469, opened 2017-06-20
Travis not failing when build - validation fails. 


camelburrito:
https://travis-ci.org/ampproject/ampstart/builds/243915998?utm_source=github_status&utm_medium=notification

camelburrito:
CC @ericlindley-g @torch2424

PR #470, opened 2017-06-21
fail when validator finds issues

erwinmombay:


erwinmombay:
Fixes https://github.com/ampproject/ampstart/issues/469

PR #474, opened 2017-06-27
Archive all

erwinmombay:
move the templates in preparation for an archive all bundle


www/render/social-share.amp.html
```diff
@@ -27,6 +27,8 @@
     {{> ../../components/socialbox/socialbox.snip.html}}
   </div>
 </main>
+
+  <script>alert(1)</script>
```
camelburrito: remove

erwinmombay: lol good catch

PR #478, opened 2017-06-28
feat(navbar): Allow the right side of the navbar to accept a link as …

WillDonohoe:
This simple pull request amends the navbar slightly to allow a link to be passed to the right side of the navbar from json, which means the nav can accept:

* A link 
* Calltracking
* Both

This allows the navbar to be slightly more flexible when dealing with different amp designs.

I wasn't sure which template to add it to for a demo as I didn't want to effect the design of them, however if you let me know which template you'd be happy for me to add it to then I can amend this PR.


WillDonohoe:
I have signed it

WillDonohoe:
I have signed it

camelburrito:
@WillDonohoe - can you give me some context into what you are trying to do? (Thanks for contributing)

WillDonohoe:
So the template design that we're implementing has a header / navbar which looks like this:

* Hamburger on the far left
* Logo in the centre
* A link on the far right

From what I can see, the area where we need to put the link only accepts calltracking, which would not be suitable in our case

gurmukhp:
This would also be useful for our template. Would it be possible to extend this to accept images too? We're looking to add a shopping cart in the top right.

components/navbar/navbar.snip.html
```diff
@@ -60,6 +60,9 @@
       {{#calltracking}}
         {{> ../calltracking/calltracking.snip.html}}
       {{/calltracking}}
+      {{#link}}
+        <a href="{{{url}}}" class="text-decoration-none px1 {{classes}}" {{#target}}target={{.}}{{/target}}>{{text}}</a>
```
camelburrito: you should not need the px1, i think either use justify-between in on the div or use consistent spacing between the elements, also please sanity check by adding both calltracking and the link at the same time (may be multiple links)

WillDonohoe:
Thanks @camelburrito, tricky one as the container expands depending on the contents so justify-between would not work in this case. I've added a margin right to each element, I've had to wrap the calltracking import in a div to add the margin.

The spacing looks good, apart from on lower resolutions. The right container is position absolute, and therefore will overlap when there's too much copy. We may need to re-look at how the layout of the navbar elements to get around this. See the below screenshots

screen shot 2017-07-06 at 5 48 43 pm

screen shot 2017-07-06 at 5 49 00 pm

camelburrito:
@WillDonohoe - looks like you have more changes on your PR than what you did (possibly due to not rebasing ) can you look into that

components/navbar/navbar.snip.html
```diff
@@ -57,9 +58,14 @@
     {{/sidebaronly}}
     {{#fixeditems}}
     <div class="ampstart-headerbar-fixed center m0 p0 mr2 flex justify-center nowrap">
-      {{#calltracking}}
-        {{> ../calltracking/calltracking.snip.html}}
-      {{/calltracking}}
+      <div class="ml1">
```
camelburrito: I think it should be mr2 - (you can drop the mr2 on the ampstart-headerbar-fixed)

WillDonohoe:
Yeah it seems the PR has been messed up as a result of rebasing. Not entirely sure of the best way to fix this, so I'm going to close this PR and re-open with a different branch

WillDonohoe:
Closing this PR in favour of [https://github.com/ampproject/ampstart/pull/494](url)

PR #481, opened 2017-06-30
Allow amp-list and amp-mustache to be used

### LINKED ISSUE #480: Unable to use amp-mustache with current workflow
- amp-mustache needs to be included on the page as a custom-template, not a custom-element. 

- amp-mustache logic is evaluated at build stage so it's not possible to use amp-list yet.

```
          <amp-list src="/img/reviews.json" width="300" height="200" layout="responsive">
            <template type="amp-mustache">
              TITLE: {{title}}
              IMG: {{imageUrl}}
            </template>
          </amp-list>
```

After build, the { characters are compiled out.

gurmukhp:
To escape mustache tags use `[[[title]]]` which will render out as `{{title}}`.

#480


camelburrito:
I think we should be using delimiters as specified here - https://mustache.github.io/mustache.5.html
(i would avoid [[]] because it could also be used in JSON.)  generally i would like to stick to mustache spec so if move to a different platform then it will not break

gurmukhp:
I investigated the custom delimiters but couldn't get it working, whenever I tried to change the default delimiter by doing:
```
// Changes delimiter from { to <%
{{=<% %>=}}
```

The `posthtml` task fails because a tag hasn't been closed.

I can manually change the delimiter to <% instead of [[[?

camelburrito:
to unblock yourself assume that this works , lets see if @erwinmombay can help

gurmukhp:
Can change it to <%:

https://github.com/gurmukhp/ampstart/commit/be5474693d720eabf8218cd6d677848907752a39

camelburrito:
Yes please

gurmukhp:
https://github.com/ampproject/ampstart/pull/482

PR #494, opened 2017-07-07
Allow the right side of the ampstart-navbar to accept a link as well as calltracking

WillDonohoe:
This simple pull request amends the navbar slightly to allow a link to be passed to the right side of the ampstart-navbar from json, which means the nav can accept:

* A link
* Calltracking
* Both

This allows the navbar to be slightly more flexible when dealing with different amp designs.

Note this PR is continuing from a previous PR found [here](https://github.com/ampproject/ampstart/pull/478)

Following from the previous PR, mr2 has been added to links, and removed from the ampstart-headerbar-fixed.

The spacing looks good, apart from on lower resolutions. The right container is position absolute, and therefore will overlap when there's too much copy. We may need to re-look at how the layout of the navbar elements to get around this. See the below screenshots

screen shot 2017-07-06 at 5 48 43 pm

screen shot 2017-07-06 at 5 49 00 pm

### DIFF
--- a/components/navbar/navbar.snip.html
+++ b/components/navbar/navbar.snip.html
@@ -58,11 +58,13 @@ limitations under the License.
     {{/sidebaronly}}
     {{#fixeditems}}
     <div class="ampstart-headerbar-fixed center m0 p0 mr2 flex justify-center nowrap">
-      {{#calltracking}}
-        {{> ../calltracking/calltracking.snip.html}}
-      {{/calltracking}}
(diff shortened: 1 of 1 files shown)


camelburrito:
The change looks good, i was waiting for you to resolve your CLA 

I guess you are using one email to commit and another email on the CLA and they both are on different accounts. 

CC @erwinmombay

WillDonohoe:
CLA issue should be resolved now

PR #497, opened 2017-07-07
Created accessible HTML Codeblocks for AMP Start UI Components

### LINKED ISSUE #393: [A11Y] Android: No appropriate way to copy the component’s html code at block level with talkback
**ENVIRONMENT:**

OS (number): N (7.0)
Build number: NRD91R
Form factor (Smartphone / tablet): Nexus 9
Browser: Chrome Dev 59.0.3062.

**PRE-REQUISITE**

1. Talkback is ON

**STEPS TO REPRODUCE:**

1. Open [https://ampstart.com](https://ampstart.com)
2. Linearly Navigate to COMPONENTS link and double tap on it.
3. Linearly navigate to any of the component’s HTML code collapsible tab and double tap on it.
4. Try to Linearly navigate and copy the HTML code.

**ACTUAL RESULTS:**

There is no way to read the code at block level. Linear navigation of code happens line by line by default. Cannot increase the granularity above beyond line level. So user can copy the code line by line only by using “Copy last utterance to clipboard” from Global Context Menu of talkback (Swipe down then right like an L shape).
Copying code Line by Line of code is fine when the html code is of 2-3 lines. But, the process becomes very redundant and time consuming when there are more lines of HTML code to be copied for the respective component.

**EXPECTED RESULTS:**

User should be able to copy the code at block level as well.

torch2424:
closes #393 

This introduces a gulp task to output escaped html files of our hl_partials folder, so that they may be included into an accesible hidden `<pre>` tag.

tested on Android talkback, and worked beautifully 😄 

# Screenshot
(Nothing is broken)
screen shot 2017-07-07 at 2 27 00 pm


torch2424:
Using the `<code>` block works just as well the the `<pre>`, Lets go with `<code>` since it is slightly more semantic in my opinion

PR #500, opened 2017-07-10
Commerce - Product details template

gurmukhp:
Hey @camelburrito,

Here is the Product Details page for the E-Commerce template.

Following our previous review, I've tried to use as many BassCSS classes as possible and created some in a similar fashion when not (ie md-h6, md-px2 etc).

There will be subsequent reviews to add the 'review' and 'related products' components. They haven't been added yet in order to keep this PR small. Future reviews should also be much smaller since many reusable components and styles will already have been defined.

As discussed, we will be using the same CSS file for all our templates with an aim to use something like uncss to remove unused code. 

**Preview Link**
https://amp-ecommerce.firebaseapp.com/templates/e-commerce/product-details.amp


css/templates/e-commerce/page-vars.css
```diff
@@ -15,12 +15,27 @@
  */
 
 :root {
+  --h1: 3.5rem;
```
gurmukhp: We went through all our designs and found we had a total of 7 header styles.

css/templates/e-commerce/page-vars.css
```diff
@@ -29,10 +44,8 @@
...
   /* Narrow custom properties */
   --narrow-page-padding: var(--space-3);
-  --narrow-nav-item-height: 3.5rem;
-  --narrow-body-font-size: .9rem;
+  --navbar-height: 3.5rem;
 
   /* Wide custom properties */
-  --wide-nav-item-height: 5rem;
-  --wide-body-font-size: 1rem;
+  --wide-navbar-height: 5rem;
```
gurmukhp: It wasn't easy to define a different navheight for mobile and desktop, so I created a new page var and redefined in main css. Whenever I tried to wrap a CSS custom property declaration around a media query, postCSS would throw an error about vars not being defined in the `:root` scope.

css/templates/e-commerce/page.css
```diff
@@ -5,51 +5,265 @@
...
+
+.mb5 {
+  margin-bottom: var(--space-6);
+}
+
+.pb5 {
+  padding-bottom: var(--space-6);
+}
+
+@media (--breakpoint-md) {
```
gurmukhp: Most headers in the design change font size from narrow to wide. There isn't an obvious correlation so setting a new font size in the body wouldn't work. Some headers stay the same size across breakpoints whilst others do not.

css/templates/e-commerce/page.css
```diff
@@ -5,51 +5,265 @@
...
+
+  .md-h6 {
+    font-size: var(--h6);
+  }
+
+  .md-h7 {
+    font-size: var(--h7);
+  }
+
+  .md-px4 {
```
gurmukhp: I've added custom padding classes akin to Bass CSS when needed. This ensured the CSS was kept small since small immutable classes are created rather than custom component names classes.

css/templates/e-commerce/page.css
```diff
@@ -5,51 +5,265 @@
...
+
+amp-selector [option] {
+  outline: 1px solid var(--color-separator);
+}
+
+amp-selector [option][selected] {
+  outline-color: var(--color-primary);
+}
+
+.amp-carousel-button {
```
gurmukhp: Disable the arrow overlays on the carousel.

css/templates/e-commerce/page.css
```diff
@@ -5,51 +5,265 @@
...
+  display: none;
+}
+
+.commerce-panel-empty {
```
gurmukhp: Some panels have a background in narrow whilst some columns have a background in wide. These classes achieve this effect.

css/templates/e-commerce/page.css
```diff
@@ -202,16 +412,78 @@ body {
...
+/* PRODUCT DETAILS */
+.icon-star,
+.icon-star-empty {
+  height: 14px;
+  width: 14px;
+  color: var(--color-font-accent);
+}
+
+.icon-star-empty {
+  color: #dadada;
```
gurmukhp: Should a new CSS var be introduced? This isn't the accent colour, should I use the same approach we did for social-color-icon?

css/templates/e-commerce/page.css
```diff
@@ -202,16 +412,78 @@ body {
...
+}
+
+@media (--breakpoint-md) {
+  .commerce-product-thumb {
+    width: 80px;
+  }
+}
+
+@media (--breakpoint-md) {
+  .commerce-product-desc {
```
gurmukhp: I couldn't find a good way to have a 1 column spacer between columns. It was either creating this margin rule or adding an empty col-1 div.

templates/e-commerce/data.json
```diff
@@ -1,13 +1,17 @@
...
+        "amp-sidebar"
+      ],
       "font-stylesheets": [
         "https://fonts.googleapis.com/css?family=Open+Sans+Condensed:300,700"
-      ]
+      ],
+      "css-path": "templates/e-commerce/page.css"
     },
     "navbar": {
-      "svg-trigger": "<svg xmlns='http://www.w3.org/2000/svg' width='24' height='24' viewBox='0 0 24 24'><path fill='none' d='M0 0h24v24H0z'/><path fill='currentColor' d='M3 18h18v-2H3v2zm0-5h18v-2H3v2zm0-7v2h18V6H3z'/></svg>",
```
gurmukhp: Removed this since the alignment in our templates wasn't correct. Need to find the correct styles in original PR and add.

templates/e-commerce/product-details.amp.html
```diff
@@ -0,0 +1,183 @@
...
+          <amp-img [src]="product[product.selectedColor].large.image2" src="{{colours.0.large.image2}}" width="1280" height="720" layout="responsive"
+            role="button" tabindex="0" alt="product image"></amp-img>
+          <amp-img [src]="product[product.selectedColor].large.image3" src="{{colours.0.large.image3}}" width="1280" height="720" layout="responsive"
+            role="button" tabindex="0" alt="product image"></amp-img>
+        </amp-carousel>
+
+        <amp-selector class="center" [selected]="product.selectedSlide" on="select:AMP.setState({product: {selectedSlide: event.targetOption}})">
+          <ul class="list-reset inline-block">
+            <li class="inline-block commerce-product-thumb">
+              <amp-img option="0" layout="responsive" selected="selected" [src]="product[product.selectedColor].thumb.image1" src="{{colours.0.thumb.image1}}"
```
gurmukhp: I didn't create a loop for images here since it's not possible in mustache to know when the first iteration of the loop is taking place without custom functions.

css/templates/e-commerce/page.css
camelburrito: http://basscss.com/#gutters

Did you try this ?

templates/e-commerce/product-details.amp.html
```diff
@@ -0,0 +1,182 @@
...
+                  <th>{{header}}</th>
+                  <td>{{body}}</td>
+                </tr>
+                {{/specs}}
+              </table>
+            </div>
+          </section>
+        </div>
+
+        <div class="reviews"></div>
```
camelburrito: do we need this ?

css/templates/e-commerce/page-vars.css
gurmukhp: Yeah it's on the high side but we did start with 11! We can review these sizes once all pages have been developed, maybe consolidating some.

css/templates/e-commerce/page.css
```diff
@@ -5,51 +5,265 @@
...
+}
+
+@media (--breakpoint-md) {
+  hr {
+    width: 100%;
+    margin: 0;
+  }
+}
+
+.commerce-select-wrapper {
```

css/templates/e-commerce/page.css
gurmukhp: I did, the spacing is only required on the right and it should be the width of one column (8.3%), however the gutters are based on rem units in whole numbers and applied to both sides. Maybe I could make the width of col-1 into a css var and then use that?

css/templates/e-commerce/page.css
```diff
@@ -5,51 +5,265 @@
...
+
+.h7 {
+  font-size: var(--h7);
+}
+
+h1, h2, h3, h4, h5, h6, h7,
+.h1, .h2, .h3, .h4, .h5, .h6, .h7 {
   line-height: normal;
-  padding-bottom: var(--space-1);
+  letter-spacing: 0;
```

css/templates/e-commerce/page.css
```diff
@@ -5,51 +5,265 @@
...
+}
+
+.pb5 {
+  padding-bottom: var(--space-6);
+}
+
+@media (--breakpoint-md) {
+  .md-h1 {
+    font-size: var(--h1);
+  }
```
gurmukhp: This is setting the font size to H1 within the medium breakpoint. In our designs there are times when the font sizes change between viewports, for example a title could be in the style of a h3 in narrow but a h1 in wide.

css/templates/e-commerce/page.css
```diff
@@ -5,51 +5,265 @@
...
+  .md-h7 {
+    font-size: var(--h7);
+  }
+
+  .md-px4 {
+    padding-left: var(--space-4);
+    padding-right: var(--space-4);
+  }
+
+  .md-pt5 {
```
gurmukhp: Nice! The responsive margin and responsive padding modules would be ideal however they are only applicable for spaces 1 -> 4 and all styles (bar 1) in the medium breakpoint are space 5+.

css/templates/e-commerce/page.css
```diff
@@ -5,51 +5,265 @@
...
+
+  .md-px7 {
+    padding-left: var(--space-7);
+    padding-right: var(--space-7);
+  }
 }
 
 
 /* ELEMENTS */
+hr {
```
gurmukhp: I couldn't find an easy way to do it using border bottom because there is some padding around each component but the separator goes full width (outside of the margins). I could redefine the padding many times for each block but felt this to be a bit overkill, what do you think? Is `<hr/>` not preferred for any reason?

css/templates/e-commerce/page.css
```diff
@@ -5,51 +5,265 @@
...
+  width: calc(100% + 2 * var(--space-3));
```

css/templates/e-commerce/page.css
```diff
@@ -5,51 +5,265 @@
...
+
+.ampstart-btn {
+  display: inline-block;
+}
+
 .ampstart-btn-secondary {
   background-color: var(--color-white);
-  color: var(--color-font-accent);
-  border-color: var(--color-font-accent);
+  color: var(--color-primary);
```
gurmukhp: This is because on the landing page, the buttons are all styled like secondary in narrow but in wider view, they are styled like a primary button. 

Maybe I could add a class like md-ampstart-btn-primary to achieve this affect? What do you think?

templates/e-commerce/product-details.amp.html
gurmukhp: It was a placeholder for future PR's. Removed.

gurmukhp:
The bundle task in the build is failing, I've raised an issue for this #502

css/templates/e-commerce/page-vars.css
camelburrito: take a look at the typography classes (also displayed on components page) and see if it makes sense to use/override them

camelburrito:
Unblocking the last comment, i have updated the bug with the cause of the issue. 

Also your build is failing because of @imports somewhere that not getting compiled.

css/templates/e-commerce/page-vars.css
gurmukhp: We could potentially use the following classes:
```
<h3 class="ampstart-title-lg">Large Title</h3>
<h3 class="ampstart-title-md">Title Text</h3>
<h3 class="ampstart-title-sm">Small Title</h3>
<h3 class="ampstart-subtitle">Subtitle text</h3>
<label class="ampstart-label">Label Text</label>
```

However they wouldn't be semantic since the subtitle would potentially be used for the title on narrow. We would also need more than 3 titles.  

I like how h1-h7 allow us to be super flexible whilst keeping the markup semantic. It's very readable.

gurmukhp:
Build is now passing. Thanks!

css/templates/e-commerce/page.css
camelburrito: nah , was just trying to see if we needed a representation in HTML to denote section break :) its still semantic, so cool to use.

css/templates/e-commerce/page.css
camelburrito: if you are not using 1 and 2 you might redefine the spacing right?

templates/e-commerce/product-details.amp.html
camelburrito: you can create a custom function if needed (in future )

css/templates/e-commerce/page-vars.css
gurmukhp: Will stick with h1-h7 for now.

css/templates/e-commerce/page-vars.css
gurmukhp: Got it.

css/templates/e-commerce/page.css
gurmukhp: Unfortunately we are using space 1 and 2. We can look to limit the spacing to 7 different values. We've done a sweep of the templates, we shouldn't need any more.

gurmukhp:
Awesome! 

All comments have been addressed, hopefully we can approve this today.

Thanks for your help!

css/templates/e-commerce/page.css
camelburrito: in general i am hesitant to do something like this , can you mark it for revision wit (using block comment), this will be good to consider in the future

PR #505, opened 2017-07-12
Nit visual changes to article template

wassgha:
### Changes
- Makes the heading's black highlight and read more button a bit more symmetric
- Gives a better initial cap letter.

Before:

<kbd>screen shot 2017-07-11 at 9 13 59 pm</kbd>

After:

<kbd>screen shot 2017-07-11 at 9 13 46 pm</kbd>

Before:

<kbd>screen shot 2017-07-11 at 9 10 02 pm</kbd>

After:

<kbd>screen shot 2017-07-11 at 8 53 50 pm</kbd>

### DIFF
--- a/components/images/fullpage-hero.css
+++ b/components/images/fullpage-hero.css
@@ -22,7 +22,7 @@
 .ampstart-image-fullpage-hero .ampstart-image-credit {
   box-decoration-break: clone;
   background: var(--color-primary);
-  padding: 0 var(--space-2) 0.2rem;
+  padding: 0.4rem;
 }
 
 .ampstart-image-fullpage-hero > amp-img {
@@ -52,10 +52,12 @@
   display: block;
   content: '⌄';
   font-size: var(--h2);
+  margin-top: 0.5rem;
 }
 
 .ampstart-readmore-text {
   background: var(--color-primary);
+  padding: 0.5rem;
 }
 
 @media (--breakpoint-md) {
--- a/css/ampstart-base/elements/initialcap.css
+++ b/css/ampstart-base/elements/initialcap.css
@@ -23,5 +23,8 @@
   color: var(--color-font-secondary);
   font-size: var(--h1);
   font-weight: var(--bold-font-weight);
-  margin-left: -2px;
+  margin-left: -7px;
+  margin-top: 7px;
+  padding: 5px;
+  float: left;
 }


camelburrito:
@wassgha - we habe a ampstart-dropcap component to get the effect you have, initial cap was intentionally like that

components/images/fullpage-hero.css
```diff
@@ -22,7 +22,7 @@
 .ampstart-image-fullpage-hero .ampstart-image-credit {
   box-decoration-break: clone;
   background: var(--color-primary);
-  padding: 0 var(--space-2) 0.2rem;
```
camelburrito: this padding was the desired padding per our original mocks , may be a good thing to run this thru @spacedino

components/images/fullpage-hero.css
```diff
@@ -52,10 +52,12 @@
   display: block;
   content: '⌄';
   font-size: var(--h2);
+  margin-top: 0.5rem;
```
camelburrito: why do we need this? we add a spacer via the navbar css if one exists.

components/images/fullpage-hero.css
```diff
@@ -52,10 +52,12 @@
...
 }
 
 .ampstart-readmore-text {
   background: var(--color-primary);
+  padding: 0.5rem;
```
camelburrito: i think this should match what is in the image credit class

css/ampstart-base/elements/initialcap.css
```diff
@@ -23,5 +23,8 @@
   color: var(--color-font-secondary);
   font-size: var(--h1);
   font-weight: var(--bold-font-weight);
-  margin-left: -2px;
```
camelburrito: This is intentional - you need to be looking at dropcap for the effect you want !

components/images/fullpage-hero.css
spacedino: @wassgha the "Read more" looks fine, so if we want the same effect for the Title Text (which we do) than the black needs to extend above the text, below the text, to the left, and right, the same amount.  Right now its 18px from the top / bottom, and 8px from the right / left of the text. Before was actually closer : 11px from the top /bottom, and 18px from the right / left. I recommend 16px top / bottom and 16px left / right padding round the text. It's divisible by 8px, which is the system our type size is built on, so it should look proportionally the best.  Does that make sense?

//cc @camelburrito also this PR name should be more descriptive...? Is this referencing an issue?

components/images/fullpage-hero.css
wassgha: This only applies to the downward-facing arrow under the "read more" to compensate for the added padding

components/images/fullpage-hero.css
camelburrito: I think @wassgha - came up with this. There is no issue on this part of the template?

wassgha: Yes, no issue. These were just the changes I made for my Show & Tell demo :)

PR #507, opened 2017-07-13
Commerce Template: Update sidebar and navbar as per latest styling

gurmukhp:
Sidebar and nav have been restyled to adhere to the design.

- Changes in colours in sidenav and navbar
- Add image in sidebar
- Add header for social icons in sidebar
- Reintroduce SVG trigger image
- Update content on landing page (text + images)


components/navbar/sidebar.snip.html
```diff
@@ -17,11 +17,14 @@
 }}
 
 <!-- Start Sidebar -->
-<amp-sidebar id="{{sidebar-id}}" class="ampstart-sidebar px3 {{^sidebaronly}}md-hide lg-hide{{/sidebaronly}}" layout="nodisplay">
+<amp-sidebar id="{{sidebar-id}}" class="ampstart-sidebar px3 {{^sidebaronly}}md-hide lg-hide{{/sidebaronly}} {{sidebarclasses}}" layout="nodisplay">
```
gurmukhp: We need this to allow us to add flexbox centering to the sidebar

components/navbar/sidebar.snip.html
```diff
@@ -17,11 +17,14 @@
...
   <div class="flex justify-start items-center ampstart-sidebar-header">
     <div role="button" aria-label="close sidebar" on="tap:{{sidebar-id}}.toggle" tabindex="0" class="ampstart-navbar-trigger items-start">✕</div>
   </div>
   <nav class="ampstart-sidebar-nav ampstart-nav">
+    {{#sidebarimage}}
```
gurmukhp: Image is to be shown before a list of links in sidebar.

components/social-follow/social-follow.snip.html
```diff
@@ -18,6 +18,9 @@
 
 
 {{#social-follow}}
+{{#header}}
```
gurmukhp: This isn't really a header which is why I've labelled it as a div.

css/templates/e-commerce/page.css
```diff
@@ -256,6 +259,22 @@ amp-selector [option][selected] {
...
+
+  .ampstart-sidebar-nav {
+    display: inline-block;
+    text-align: center;
+  }
+}
+
+.ampstart-sidebar-nav-image {
+  width: 50%;
+  min-width: 150px;
```
gurmukhp: min width of image should be in px due to the size of the source image.

css/templates/e-commerce/page.css
```diff
@@ -265,14 +284,23 @@ amp-selector [option][selected] {
   z-index: var(--z1);
 }
 
+@media (--breakpoint-md) {
+  .ampstart-sidebar .ampstart-sidebar-header {
```
camelburrito: just  .ampstart-sidebar-header should work ?

components/navbar/sidebar.snip.html
camelburrito: one small nit 
can you do 
```json
sidebar: {
  "classes": "general purpose classes",
  "layout-classes": "ablsolute, relative, float etc",
  "spacing-classes": "p0, m0, mx-auto etc",
  "resposive-display-classes": "sm-hide md-hide",
  "resposive-layout-classes": "md-col12, sm-col6 etc"

}
```

I think this is becoming a pattern now ,  when we build the next version of ampstart , it will help separate markup and layout very easily.

components/social-follow/social-follow.snip.html
camelburrito: is this the title for the section? if so can we add it on the page?

components/navbar/sidebar.snip.html
gurmukhp: Does this mean we shouldn't hardcode any classes into the template itself? Changing this file would require changing almost all pages within ampstart (at least 16), would it be possible to file this as a bug and do it in a follow up PR?

components/navbar/sidebar.snip.html
gurmukhp: Done.

components/social-follow/social-follow.snip.html
gurmukhp: If we were including social-follow directly in our templates, we could. However social-follow is included by sidebar which is included by navbar. I could add it to the sidebar component, but I don't see how that is any different to adding it here.

Which file should I add the title to?

css/templates/e-commerce/page.css

gurmukhp:
Comments addressed. Thanks!

components/navbar/sidebar.snip.html
camelburrito: I am going to write a design doc to come up with an idea to separate markup , layout and spacing so all of this is easily configurable. Since i saw a lot of PRs adding classes i thought it might be easy to segregate this was, because my design will take a while to be ready for implementation. However this is not very imp at this point.

components/social-follow/social-follow.snip.html
camelburrito: You are correct , The reason i wanted this to stay in the template (the respective page) is to easily add a heading level (You might also need a wrapping div to keep the component markup in a division). I think this should be a heading. You can check for other files on how we passing heading level via a config. 
i think default level could be h3

PR #508, opened 2017-07-13
Commerce Template: Product listing page

gurmukhp:
Product listing page for Commerce template.

Loads products from a static JSON endpoint which is generated using a simple node script.

**Preview**
https://amp-ecommerce.firebaseapp.com/templates/e-commerce/product-listing.amp


components/inputs/input-vars.css
```diff
@@ -20,4 +20,5 @@
   --color-input-font-primary: var(--color-font-primary);
   --color-input-secondary: var(--color-secondary);
   --color-input-font-secondary: var(--color-font-secondary);
+  --input-checkbox-radio-size: 20px;
```
gurmukhp: Set the input size of checkbox and radio as a css var so we can easily override this in our template.

components/inputs/inputs.snip.html
```diff
@@ -78,7 +78,7 @@
 
 {{#radio}}
 <!-- Start Radio -->
-<div class="ampstart-input ampstart-input-radio inline-block relative m0 p0 mb3 {{classes}}">
+<div class="ampstart-input ampstart-input-radio inline-block relative m0 p0 mb3 {{classes}}" {{#on}}on="{{.}}"{{/on}}>
```
gurmukhp: by defining 'on' here, we're able to use amp-bind easily.

css/templates/e-commerce/page.css
```diff
@@ -1,5 +1,6 @@
 @import '../../core';
 @import 'basscss-responsive-type-scale';
+@import 'basscss-responsive-margin';
```
gurmukhp: Add responsive addon so we're able to use md-m0 -> md-m4

css/templates/e-commerce/page.css
```diff
@@ -95,15 +100,14 @@ dd:after {
 
 /* TODO: Merge with existing amp-start select component */
 .commerce-select-wrapper {
-  display: inline-block;
   position: relative;
-  padding-right: 20px;
+  padding-right: 13px;
```
gurmukhp: Spacing to allow for down arrow in commerce-select. This shouldn't really be configurable and is used once, hence not a css var.

css/templates/e-commerce/page.css
```diff
@@ -448,3 +467,21 @@ amp-selector [option][selected] {
...
+
+/* PRODUCT LISTING */
+.commerce-listing-banner {
+  width: 100%;
+}
+
+@media (--breakpoint-md) {
+  .commerce-listing-header {
+    margin-top: var(--space-3);
+    border-top: 1px solid var(--color-separator);
```
gurmukhp: Able to have separators as borders here.

css/templates/e-commerce/page.css
```diff
@@ -448,3 +467,21 @@ amp-selector [option][selected] {
...
+    border-bottom: 1px solid var(--color-separator);
+  }
+
+  .commerce-listing-wrapper {
+    margin-left: calc(1/12 * 100%);
```
gurmukhp: Had to find a way to use up the space of 1 column. Gutters don't seem to do what's needed. I've defined this declaration twice, should I make a --col-1 css var?

gulpfile.js
```diff
@@ -130,6 +130,10 @@ gulp.task('img', function() {
   return gulp.src(config.src.img).pipe(gulp.dest(config.dest.img));
 });
 
+gulp.task('templatedata', function() {
+  return gulp.src(config.src.templateData).pipe(gulp.dest(config.dest.templateData));
```
gurmukhp: JSON endpoints containing products are stored in templates/data/*.json. This should be copied over to dist.

templates/e-commerce/data/build.js
```diff
@@ -0,0 +1,107 @@
+const fs = require('fs');
```
gurmukhp: This is a simple node JS script that generates some products for us to use on the listings page. It creates all endpoints for 4 categories with two filters (price high to low and vice versa). 

Not sure if this file belongs here or just the output files. What do you think?

templates/e-commerce/data/high-low-all-products.json
```diff
@@ -0,0 +1,88 @@
+{
```
gurmukhp: Sample JSON endpoints to show filtered products.

css/templates/e-commerce/page.css
```diff
@@ -120,6 +124,16 @@ dd:after {
...
 }
 
+.ampstart-input-radio {
+  display: block;
+  margin-bottom: var(--space-1);
+}
+
+.ampstart-input-radio label {
+  font-size: var(--h6);
+  cursor: pointer;
```
camelburrito: this should be the default you can move it up?

gulpfile.js
camelburrito: instead of calling them data - i guess you could give it a different foldername like rest/ or api/ etc, but the idea sounds good to me.

templates/e-commerce/data/build.js
camelburrito: i think the output files need to be in this dir, i was expecting the products could be static , i guess you can add this to the tasks/e-commerce folder?. Let me know what you think

templates/e-commerce/product-listing.amp.html
```diff
@@ -0,0 +1,74 @@
...
+        {{#price}} {{> partials/select.snip.html}} {{/price}}
+      </div>
+
+      <amp-list class="mx1 md-mxn1" [src]="'data/' + products.filter + '-' + products.category + '-products.json'" src="data/high-low-all-products.json" height="500" width="300" layout="responsive">
+        <template type="amp-mustache">
+          <div class="commerce-listing-product-item inline-block col-6 md-col-3 px1 mb2 md-mb4">
+            <amp-img src="<%image%>" width="340" height="340" layout="responsive" alt="<% name %>"></amp-img>
+            <h2 class="h6"><% name %></h2>
+            <% description %>
+            <div class="h6">&pound;<% price %></div>
```
camelburrito: super nit: if you get the right utf8 char , you can add it instead of `&pound;`

gulpfile.js
gurmukhp: Renamed to `api`

templates/e-commerce/data/build.js
gurmukhp: The products will definitely be static.

This script was built to help generate a product library quickly, we are going to update the product images, prices etc as time goes on. So I could omit this file and keep it as a public gist for others to use if they need to? What do you think?

There is no need for the products to get generated everytime we run a gulp build.

templates/e-commerce/product-listing.amp.html
gurmukhp: Updated to `&#163;`, why is this preferred over a named entity?

css/templates/e-commerce/page.css

PR #514, opened 2017-07-15
Undo Firebase functions revert

camelburrito:


gurmukhp:
Looks good! Thanks for reverting.

PR #516, opened 2017-07-17
[Commerce] Blog Listing & Blog

gurmukhp:
Commerce Blog Listing and Blog template.

- Mocks added as per design doc

**Preview**
https://amp-ecommerce.firebaseapp.com/templates/e-commerce/blog-listing.amp


css/templates/e-commerce/page.css
```diff
@@ -493,15 +522,29 @@ amp-selector [option][selected] {
...
+}
+
+@media (--breakpoint-md) {
+  .commerce-blog {
+    margin-left: calc(1/12 * 100%);
+  }
+}
+
+.commerce-blog p {
+  line-height: 1.6;
```
gurmukhp: I don't think this should be a css var since it's only used once and shouldn't really be configured.

css/templates/e-commerce/page.css
```diff
@@ -493,15 +522,29 @@ amp-selector [option][selected] {
...
 
 
+/* BLOG */
+.commerce-blog-wrapper {
+  background-color: var(--color-white);
+}
+
+@media (--breakpoint-md) {
+  .commerce-blog {
+    margin-left: calc(1/12 * 100%);
```
gurmukhp: I'm using this in 3 places now. Should I make it a css-var or make a utility class?

templates/e-commerce/blog-article.amp.html
```diff
@@ -0,0 +1,55 @@
...
+<body>
+  {{> ../../components/navbar/navbar.snip.html}} {{/header}}
+
+  <main id="content" role="main" class="main flex flex-wrap commerce-panel-filled">
+    {{#blog.articles.0}}
+    <article class="col-12 md-col-8 commerce-blog-wrapper pt2 md-pt4">
+      <amp-img class="md-col-10" src="{{images.0.narrow}}" srcset="{{images.0.narrow}} 750w,{{images.0.wide}} 1280w" width="{{imageWidth}}"
+        height="{{imageHeight}}" layout="responsive" alt="{{title}}"></amp-img>
+      <div class="md-col-10">
+        <div class="px2 mt3">
```
camelburrito: i think the 2 divs can be merged into one.

templates/e-commerce/blog-article.amp.html
```diff
@@ -0,0 +1,55 @@
...
+          <time datetime="{{datetime}}" class="block mb2">{{date}}</time>
+          <p class="mb5">{{paragraphs.0}}</p>
+        </div>
+      </div>
+      <amp-img class="mb3 md-col-10" src="{{images.1.narrow}}" srcset="{{images.1.narrow}} 750w,{{images.1.wide}} 1280w" width="{{imageWidth}}"
+        height="{{imageHeight}}" layout="responsive" alt="{{title}}"></amp-img>
+      <amp-img class="mb5 md-col-10" src="{{images.2.narrow}}" srcset="{{images.2.narrow}} 750w,{{images.2.wide}} 1280w" width="{{imageWidth}}"
+        height="{{imageHeight}}" layout="responsive" alt="{{title}}"></amp-img>
+      <div class="md-col-10">
+        <div class="px2 mb4">
```
camelburrito: ditto

templates/e-commerce/blog-article.amp.html
```diff
@@ -0,0 +1,55 @@
...
+          <p>{{paragraphs.1}}</p>
+          <blockquote class="commerce-blockquote md-col-8">{{quote.text}}</blockquote>
+          <div class="commerce-blockquote-quoter">- {{quote.quoter}}</div>
```
camelburrito: is this different from ampstart-blockquote? 
I mean not visually - functionally is this different?
If not we would need to rewrite the ampstart-blockquote on your page.

templates/e-commerce/blog-listing.amp.html
```diff
@@ -0,0 +1,35 @@
...
+-->
+
+<html ⚡>
+{{#header}} {{> ../../utils/amp-page-head.snip.html}}
+<body>
+  {{> ../../components/navbar/navbar.snip.html}} {{/header}}
+
+  <main id="content" role="main" class="main commerce-panel-filled">
+    <section class="col-12 md-col-8 commerce-blog-wrapper pt2 md-pt1">
+        <div class="xs-hide sm-hide md-commerce-header md-mb3 md-col-10">
```
camelburrito: cant we move all these classes to the H1?

templates/e-commerce/partials/blog-preview.html
```diff
@@ -0,0 +1,11 @@
+{{#articles}}
+<article class="md-pb7">
+  <amp-img class="mb3 md-col-10" src="{{images.0.narrow}}" srcset="{{images.0.narrow}} 750w,{{images.0.wide}} 1280w" width="{{imageWidth}}" height="{{imageHeight}}" layout="responsive" alt="{{title}}"></amp-img>
```
camelburrito: why do we use 'images.0' (i also saw this some other places)?

templates/e-commerce/blog-article.amp.html
gurmukhp: Done.

templates/e-commerce/blog-article.amp.html
gurmukhp: Done, thanks for catching.

templates/e-commerce/blog-article.amp.html
gurmukhp: Created pullquote ampstart component and inherited it.

templates/e-commerce/blog-listing.amp.html
gurmukhp: Merged.

templates/e-commerce/partials/blog-preview.html
gurmukhp: This is to target only the first element in the array. I'm using the same 'blog' object for both the blog listings and blog article. Only the first blog has the full article data in it. Therefore by using `.0` we are able to target it.

camelburrito: if narrow and wide are set on all elements in the array, you should not have to worry about this right? 
It is not great to rely on something like this here because of some one replaces your first image they might not expect this consequence. 

```json
{
  "articles": [
    {
       "image": {
         "narrow": "",
         "wide": "",
         "title": ""
        }
    }
  ]
}
```

gurmukhp: Specifically `images.0.narrow` is used here because only the first image is needed to show the preview. 

This approach is used in the blog-article template. Only the first blog data is needed and all 3 images are mandatory to build the page. I could do some error checking to ensure the image exists before displaying it, would that solve this problem?

css/templates/e-commerce/page.css
```diff
@@ -182,6 +196,21 @@ amp-selector [option][selected] {
...
 
+.ampstart-pullquote {
+  font-size: var(--h2);
+  margin: var(--space-3) 0 var(--space-2) 0;
+  line-height: 1.2;
+  border-left: none;
+  padding-left: 0;
+}
+
+.commerce-pullquote-quoter {
```
camelburrito: @gurmukhp 
nit: author? - you can change in the next PR

PR #518, opened 2017-07-17
[LUNE] - Landing page

WillDonohoe:
Pull request for the landing page of the LUNE template.

Please see the mocks in the mocks/ folder. 

We've refactored a lot of the code to fit with bass css, however this page especially comes with some fairly bespoke functionality for button animations and scrolling animation effect so custom styles were needed for these.

See below for the staging link:
https://gweb-amp-template-lune-review.firebaseapp.com/templates/lune/lune.amp

Thanks


css/templates/lune/_type.css
```diff
@@ -0,0 +1,31 @@
...
+  TODO: Raise issue with base css to get this fixed. */
+
+.h00 {
+  font-size: var(--h00);
+}
+
+.h0 {
+  font-size: var(--h0);
+}
+
```
WillDonohoe: Responsive typescale defines sm-h0, sm-h00, but does not define .h0 for viewports smaller than 640px. So we've added them in here

css/templates/lune/components/_cta.css
```diff
@@ -1,3 +1,117 @@
...
+.lune-theme-dark .lune-component-cta:focus > span {
+  color: var(--color-background-dark-narrow);
+}
+
+@media (--breakpoint-mt) {
+  .lune-theme-dark .lune-component-cta:hover > span,
+  .lune-theme-dark .lune-component-cta:focus > span {
+    color: var(--color-background-dark-wide);
+  }
+}
```
WillDonohoe: `lune-component-cta` inherits from `ampstart-button`, with some added bespoke animations on hover / focus

css/templates/lune/page-vars.css
```diff
@@ -62,17 +63,22 @@
...
+  --animation-delay-step: .1s;
+
   /*
    * Copying these @custom-media rules and the --space-* variables from
-   * bass-override-vars.css. This is a temporary solution, until we merge
-   * all the CSS into one file.
+   * bass-override-vars.css. This is a temporary solution, waiting for a
+   * better way of doing it.
    */
   @custom-media --breakpoint-xs (max-width: 40rem);
```
camelburrito: not sure why you have to do this , it should work by default without needing to add this here again.

css/templates/lune/pages/_landing.css
```diff
@@ -1,3 +1,152 @@
...
+    }
+
+    .lune-landing-wrapper {
+      position: fixed;
+    }
+  }
+}
+
+/*
+ * Because amp-img can't be resized in a responsive way through changing its height,
```
camelburrito: have you tried the heights attribute? 

https://www.ampproject.org/docs/guides/responsive/art_direction#heights

templates/lune/lune.amp.html
```diff
@@ -18,6 +18,69 @@
   {{> ../../utils/amp-page-head.snip.html}}
   <body class="lune">
     {{> ../../components/navbar/navbar.snip.html}}
+
+    <!-- Used to achieve the clipping effect when scrolling the page -->
+    <div class="hide">
```
camelburrito: can you attach a small gif of what this does in the comment.

templates/lune/lune.amp.html
```diff
@@ -18,6 +18,69 @@
...
+                  <!-- WebP image -->
+                  <amp-img
+                    class="lune-anim-scale-up z1"
+                    alt="{{alt}}"
+                    srcset="{{webp_srcset}}"
+                    width="{{width}}"
+                    height="{{height}}"
+                    layout="responsive">
+                      <!-- PNG Fallback -->
+                      <amp-img
```
camelburrito: I guess you could do a fallback svg? you might avoid the extra nextwork call.

css/templates/lune/_type.css
WillDonohoe: Yeah, check out the responsive type scale add on. Code here:
https://github.com/basscss/addons/blob/master/modules/responsive-type-scale

You'll notice inside `index.css`, `h0` and `h00` are defined. And inside `lib/sm-type-scale.css` the `.sm-h0` and `.sm-h00` classes are defined, but they sit within the `--breakpoint-sm` breakpoint, so we can't use these sizes on smaller devices.

templates/lune/lune.amp.html
WillDonohoe: This fallback is intended for browsers that do not support webp, so an SVG would not be suitable in this case. We're following the example found [on this page](https://www.ampproject.org/docs/reference/components/amp-img) Go to the "Example: Specifying a fallback image" section

templates/lune/lune.amp.html
ciampo: You can find the gif for the effect we need to achieve at `mocks/lune/landing-page-clipping.gif`.

In particular, this `<div class="hide">`  contains the `svg` that is used in the `clip-path` CSS property (see `css/templates/lune/pages/_landing.css`(

css/templates/lune/pages/_landing.css
```diff
@@ -1,3 +1,157 @@
...
+
+  .lune-landing-section.lune-theme-dark {
+    position: fixed;
+  }
+
+  /*
+   * Only apply this enhancement if clip-path is supported
+   * and the viewport is high enough.
+   */
+  @supports (clip-path: inset(1px 0)) {
```
ciampo: Progressive enhancement applied only when `clip-path` is supported.

Note that in Chrome the result is a bit buggy — a ticket has been raised here https://bugs.chromium.org/p/chromium/issues/detail?id=740962

css/templates/lune/pages/_landing.css
ciampo: I tried using `heights` (and also `layout="fill"`), but I couldn't really achieve the responsive behavior I need, unless I used the `object-fit` CSS property.
Although the [docs](https://ampbyexample.com/advanced/how_to_support_images_with_unknown_dimensions/#object-fit-to-the-rescue) encourage the use of `object-fit`, we can't really use it as its browser support isn't good enough (no IE, no edge).

I believe that the current solution is a good one, as it fulfills completely the design requirements and works across all major browsers

templates/lune/lune.amp.html
camelburrito: I meant add it in the code comments :)

css/templates/lune/page-vars.css
```diff
@@ -1,6 +1,7 @@
 :root {
   --content-max-width: 100%;
 
+  --z0: 0;
```
camelburrito: same here - you wll not need all the z-indices that were already defined. (You will only need --z0)

css/templates/lune/pages/_landing.css
```diff
@@ -1,3 +1,166 @@
...
+.lune-landing-section {
+  height: var(--lune-landing-page-height);
+  min-height: var(--lune-landing-page-min-height);
+}
+
+.lune-landing-wrapper {
+  top: var(--navbar-height);
+}
+
+.lune-page-landing .lune-component-cta {
```
camelburrito: should this be ampstart-btn?

css/templates/lune/pages/_landing.css
```diff
@@ -1,3 +1,166 @@
...
+  animation-delay: calc(var(--animation-delay-start) +
+      var(--animation-delay-step));
+}
+
+.lune-landing-title-bottom {
+  animation-delay: calc(var(--animation-delay-start) +
+      2 * var(--animation-delay-step));
+}
+
+.lune-page-landing .lune-component-cta {
```
camelburrito: should this be ampstart-btn

css/templates/lune/pages/_landing.css
camelburrito: you should be using `layout="responsive"` look at how we did the fullpage hero using 2 images and see if it can help you!

ciampo: If you refer to the [full hero component](https://ampstart.com/render/fullpage-hero.amp), it's using `object-fit` too.  It won't work in IE or Edge, as the image will be stretched
screen shot 2017-07-18 at 17 39 46

ciampo:
We're happy for this PR to be closed in favour of #518

PR #520, opened 2017-07-18
[Commerce] Cart

gurmukhp:
- Shopping cart for commerce template
- mocks added


css/templates/e-commerce/page.css
```diff
@@ -557,3 +557,32 @@ amp-selector [option][selected] {
...
+  .commerce-cart-item {
+    border-bottom: 1px solid var(--color-separator);
+  }
+
+  .commerce-cart-button {
+    text-align: left;
+  }
+
+  .commerce-cart-item-image {
+    max-width: 100px;
```
gurmukhp: Prevent image from getting larger than 100px on wider screens since this is inside of a `col-2`. 

The only other way I could think of doing this was to have two `amp-img`, one with fixed layout and another with responsive and then show/hide dependant on the device width. I did't do this approach since it feels quite hacky.

css/templates/e-commerce/page.css
```diff
@@ -557,3 +557,32 @@ amp-selector [option][selected] {
...
+  border-bottom: 1px solid var(--color-separator);
+}
+
+@media (--breakpoint-md) {
+  .commerce-cart-item {
+    border-bottom: 1px solid var(--color-separator);
+  }
+
+  .commerce-cart-button {
+    text-align: left;
```
gurmukhp: Align button to left on wider screens. I could avoid this div but it'd mean having to change the layout of the parent container to flex. The downside of this approach is the button width wouldn't be determined by the button text (unless I add an empty div around the button), which puts us in the same place (in terms of markup).

css/templates/e-commerce/page.css
```diff
@@ -557,3 +557,32 @@ amp-selector [option][selected] {
...
+  }
+
+  .commerce-cart-item-image {
+    max-width: 100px;
+  }
+
+  .commerce-cart-summary {
+    height: calc(100vh - var(--wide-navbar-height));
```
gurmukhp: Ensure cart summary section on right side of wider devices is as tall as the page.

gurmukhp:
Merge conflicts fixed.

Issue #526, opened 2017-07-19
[navbar] clicking on a svg-trigger icon doesn't open the sidebar on IE11


ciampo:
When using the `svg-trigger` option to provide a custom `svg` icon for the navbar, the sidebar won't open on IE11 when clicking the icon.

This issue can be reproduced by opening both the [e-commerce](https://gweb-amp-template-lune-review.firebaseapp.com/templates/e-commerce/landing.amp) and the [lune](https://gweb-amp-template-lune-review.firebaseapp.com/templates/lune/product-blanc.amp) templates in IE11. When clicking on the hamburger icon, the sidebar doesn't open.

A possible fix is to set `pointer-events: none` on the `svg` element; in that case, the `div[role="button"]` element that wraps the `svg` can correctly receive tap events and toggle the sidebar.

PR #528, opened 2017-07-19
[navbar] fix click events on sidebar toggle when using a custom svg-icon in IE11

ciampo:
See related issue #526 .

Setting `pointer-events: none` allows the click event to propagate to its wrapper and toggle the sidebar.

### DIFF
--- a/components/navbar/navbar.css
+++ b/components/navbar/navbar.css
@@ -137,3 +137,7 @@
 .ampstart-sidebar .ampstart-navbar-trigger {
   line-height: inherit;
 }
+
+.ampstart-headerbar .ampstart-navbar-trigger svg {
+  pointer-events: none;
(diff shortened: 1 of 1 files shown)


components/navbar/navbar.css
```diff
@@ -137,3 +137,7 @@
 .ampstart-sidebar .ampstart-navbar-trigger {
   line-height: inherit;
 }
+
+.ampstart-headerbar .ampstart-navbar-trigger svg {
```
camelburrito: Just `.ampstart-navbar-trigger svg` should be fine. is this because the svg does not bubble up click events?

ciampo: Yes, the `svg` elements swallows the click events, which never bubble up

PR #534, opened 2017-07-20
Removed Redundant Dependencies, and Created A Count CSS Task

torch2424:
- Removed the dependencies `gulp-intercept` and `gulp-ext-rename`
- Created a task to count the number of characters in css with dist/ and output it's size in both number of characters, and bytes

cc @camelburrito are there any relevant issues for this? Also, let me know if this should be output to dist/ or src/ since I know you wanted to use this as a PR tool.

Example Output:

````
File name: core.css
Number of characters: 2139
Size: 2.139 KB
````

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -116,7 +116,7 @@ function getPartials(acc, embedderDir, template) {
 
 gulp.task('build', 'build', function(cb) {
   runSequence(
-      'clean', 'highlight', 'escape', 'img', 'templateapi', 'postcss', 'posthtml', 'www', 'validate',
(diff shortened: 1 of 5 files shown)


camelburrito:
Can you checkin the size.txt?

torch2424:
Also, I just included the filename in the output.

@camelburrito so you would like these files output in src on build then?

camelburrito:
cc @erwinmombay

camelburrito:
@torch2424 - This was not the requirement , lets sync on this tomorrow and finish this.

PR #535, opened 2017-07-20
[Commerce] Use amp-bind to update main class

gurmukhp:
I've moved the `[class]` binding to the main element. This ensures we don't have to write anything hacky in the JSON files or the templates. It also ensures that we can keep the logic for the interaction within the same template.


templates/e-commerce/product-details.amp.html
```diff
@@ -109,7 +116,11 @@ <h2 class="h7 block pb1">Reviews</h2>
         </div>
         <hr class="mb4" />
         <div class="col-12 self-start mb4 commerce-product-btn-wrapper">
+<<<<<<< HEAD
```
camelburrito: merge conflicts ?

gurmukhp: Sorry I'm not sure why I'm not seeing these, will investigate.

Fixed

PR #543, opened 2017-07-21
[LUNE] - Landing page QA

ciampo:
This PR contains a few bugfixes to the landing page, in particular to the clipping effect.

We are now using both the old `clip` property and the more recent `clip-path` property, plus a couple of changes on positioning. These changes fix a couple of cross-browser bugs related mainly to rendering glitches and pointer events.

### DIFF
--- a/css/templates/lune/pages/_landing.css
+++ b/css/templates/lune/pages/_landing.css
@@ -118,6 +118,11 @@
    */
   @supports (clip-path: inset(1px 0)) {
     .lune-landing-section.lune-theme-light {
+      /*
+       * Using the old clip spec too, as it works better in some browser.
+       * Requires the element to have position:absolute.
+       */
+      clip: rect(0 auto auto 0);
       clip-path: url(#svgClipPath);
       /*
        * We need the height to be a bit more than its normal value (hence the 0.99 factor),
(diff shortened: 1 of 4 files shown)


templates/lune/lune.amp.json
```diff
@@ -10,7 +10,7 @@
 
   "landing-sections": [
     {
-      "classes": "z2 relative lune-theme-light",
+      "classes": "z2 absolute lune-theme-light",
```
ciampo: `position: absolute` is required by the CSS `clip` spec

css/templates/lune/pages/_landing.css
```diff
@@ -127,8 +132,18 @@
       height: calc(var(--lune-landing-page-height) / 0.99);
     }
 
-    .lune-landing-wrapper {
+    .lune-landing-content {
```
ciampo: Setting `position: fixed` on `.lune-landing-wrapper` was causing a few bugs related to pointer events. To avoid these bugs, we have to set `position: fixed` on `.lune-landing-content` and `.lune-component-scroll-indicator` instead, and thus the only way to center them in the page is by using `transform`.

This may not be very flexible, but it's the only way of achieving the design, which is very bespoke for this page.

camelburrito: one minor nit - can the fixed be moved into the HTML ?

ciampo: Not really, as the position fixed is applied inside a @support query. If the browser doesn't support clip-path, those element have another value for their position

PR #546, opened 2017-07-22
Undoing Functions revert for the second time

camelburrito:
Hopefully this does not get deleted again

CC @ciampo @gurmukhp


camelburrito:
@erwinmombay can you LGTM pls

PR #547, opened 2017-07-23
Strip !Important and replace it with :root:not(#id) hack

camelburrito:
Strip !Important and replace it with :root:not(#id) hack
Specificity will be atleast 100 more than the greatest specificity in the document's CSS


camelburrito:
I would like to convert this to an NPM Package. And use it across the board.

camelburrito:
Closing this PR - we have merged this logic in tools/

PR #549, opened 2017-07-24
[Commerce] Add sidebar to blog templates

gurmukhp:


css/templates/e-commerce/page.css
```diff
@@ -590,6 +591,10 @@ amp-carousel {
   line-height: 1.6;
 }
 
+.commerce-blog-sidebar .ampstart-social-follow li:first-child a {
```
gurmukhp: This is quite a long selector but I couldn't find an easy way to do this apart from adding a class via json, but it wouldnt adhere to the design doc naming style, so would therefore need to change all templates inheriting this component.

templates/e-commerce/partials/blog-sidebar.snip.html
```diff
@@ -0,0 +1,8 @@
+<aside class="xs-hide sm-hide md-col-4 md-px4 md-pt6 commerce-blog-sidebar">
+  <h1 class="h5 md-pb2">The bike shop blog</h1>
+  <p class="mb2">Praesent commodo cursus magna, vel scelerisque nisl consectetur et. Duis mollis, est non commodo luctus, nisi erat porttitor ligula, eget lacinia odio sem nec elit. Aenean lacinia bibendum nulla sed consectetur. Aenean eu leo quam. Pellentesque ornare sem lacinia quam venenatis vestibulum. Praesent commodo cursus magna, vel scelerisque nisl consectetur et. </p>
+  {{#header.navbar}}
+    {{> ../../../components/social-follow/social-follow.snip.html}}
+  {{/header.navbar}}
+  <amp-ad layout="responsive" width="300" height="250" type="industrybrains" data-width="300" data-height="250" data-cid="19626-3798936394" class="mt2"/>
```
gurmukhp: I copied this example from Ampproject. The ad displays fine but there is a cross origin error. Is there a different ad we could use that doesn't have any errors?

camelburrito: checkout - examples folder on amphtml repo and see if it works

gurmukhp: Thanks, the thin mobile doubleclick ad worked, but some of the others return "server at capacity" whilst others just failed to load. I'll leave this one in for now as it renders at least.

PR #550, opened 2017-07-24
[Commerce] Animate image on hover for hero section

gurmukhp:
- Align text as per max width
- Move footers to thinner columns


templates/e-commerce/partials/hero.snip.html
```diff
@@ -1,18 +1,20 @@
...
-    <div>
-      <h2 class="commerce-hero-content-title h4 md-h1 caps">{{title}}</h2>
-      <p class="commerce-hero-content-body">{{body}}</p>
+  <div class="commerce-hero-content commerce-hero-content-theme-{{text-theme}} flex justify-center flex-column md-mx-auto md-p0">
+    <div class="commerce-hero-align md-mx-auto md-pl4">
+      <h2 class="commerce-hero-content-title h4 md-h1 caps md-mb2">{{title}}</h2>
+      <p class="commerce-hero-content-body md-mb4">{{body}}</p>
       {{#link}}
-        <a href="{{url}}" class="ampstart-btn ampstart-btn-secondary caps">{{text}}</a>
+        <div class="ampstart-btn ampstart-btn-secondary caps">{{text}}</div>
```
gurmukhp: Changing this to a div since the outer container is a link and it's not valid html to have a link inside a link.

templates/e-commerce/partials/hero.snip.html
```diff
@@ -1,18 +1,20 @@
...
```
camelburrito: can you use a button instead?

gurmukhp: As discussed, will leave as a div since this will load a new page rather than interact on a page.

PR #551, opened 2017-07-24
[Commerce] Align hero text to max width

gurmukhp:


css/templates/e-commerce/page.css
```diff
@@ -505,6 +505,8 @@ amp-selector [option][selected] {
   .commerce-hero-content-body {
     font-size: var(--h6);
     max-width: none;
+    margin-bottom: var(--space-4);
```
gurmukhp: Unable to add these as basscss classes due to an override in a previous class.

PR #553, opened 2017-07-25
Created a Browserify Gulp Task for the AMP Start Configurator

torch2424:
* Include Browserify into the project, as well as some `preety-bytes` for testing
* Creates a simple test page for testing our configurator tasks
* Creates a gulp for bundling JS using Browserify
* Creates a gulp task for copying static files from configurator/src to the build directory

# Screenshot
screen shot 2017-07-24 at 5 31 56 pm


torch2424:
Decided to use webpack after chatting with @erwinmombay

PR #558, opened 2017-07-25
[LUNE] - Materials pages

WillDonohoe:
This PR contains the final pages for the LUNE template. 2 pages with different content, but follow the same layout.

Mocks included in the /mocks/ directory:
Wide - Materials 1.jpg
Wide - Materials 2.jpg
Narrow - Materials 1.jpg
Narrow - Materials 2.jpg


templates/lune/partials/product/overlay.snip.html
```diff
@@ -0,0 +1,208 @@
...
+                  alt="{{alt}}"
+                  srcset="{{fallback_srcset}}"
+                  layout="responsive"
+                  width="{{width}}"
+                  height="{{height}}"></amp-img>
+              </amp-img>
+            {{/desktop}}
+          </figure>
+        {{/image}}
+        <div class="col-1 flex-none xs-hide sm-hide"></div>
```
camelburrito: what is this div for?

WillDonohoe: To mimimise the amount of custom css, we are using divs here with bass css columns to create gutters. The alternative is to use custom css to create margins with the same column % widths.

Issue #432, opened 2017-05-12
Use native sticky header in at least one theme


ericlindley-g:
More a reminder than anything else, but a cool thing to feature to highlight support in AMP

PR #560, opened 2017-07-26
[Commerce] Sticky sidebar

gurmukhp:
Ensure sidepanels on wider viewports are fixed to the top whilst the page scrolls.

Potentially a fix for #432 ?


css/templates/e-commerce/page.css
```diff
@@ -216,7 +216,8 @@ amp-selector [option][selected] {
 
 @media (--breakpoint-md) {
   .commerce-side-panel {
-    background-color: var(--color-panel);
+    position: sticky;
```
camelburrito: this might not work in IE - but might be okay if the experience is decent on IE

gurmukhp:
Merge conflicts addressed.

css/templates/e-commerce/page.css
gurmukhp: This style ensures that the side panel stays fixed as the user scrolls. If sticky is not supported, the panel will scroll with the rest of the page. So it degrades gracefully :)

PR #566, opened 2017-07-27
[Commerce] Creative Review Bugs

raisaveuc:
* Wrap cart contents to prevent overflow of fixed side panel when too
short
* Fix sticky top position calculation bug
* Remove sticky style to products sidebar where content is too long
* Cleanup unused styles and elements
* Move negative footer margins to outer element
* Add drop-shadow and button margin to cart notification
* Move responsive spacing classes for naming accuracy
* Add sticky positon to categories filter
* Fix blocking cart notification
* Make cart notification animation quicker


templates/e-commerce/cart.amp.html
```diff
@@ -21,11 +21,13 @@
...
   {{> ../../components/navbar/navbar.snip.html}} {{/header}}
-  <main id="content" role="main" class="main flex flex-wrap items-start commerce-cart pb3 md-pb7 md-mb3">
-    <section class="col-12 md-col-8 pt2 pb3 md-pt4 md-pb7">
-      {{> partials/cart-items.snip.html}}
-    </section>
-    {{> partials/cart-summary.snip.html}}
+  <main id="content" role="main" class="main items-start commerce-cart pb3 md-pb7 md-mb3">
+    <div class="flex flex-wrap">
+      <section class="col-12 md-col-8 pt2 pb3 md-pt4 md-pb7">
+        {{> partials/cart-items.snip.html}}
```
camelburrito: you should probably move the section inside the partial?

raisaveuc: carrt-items.snip.html is used across 3 templates in very different containers. I took a look again at the templates using this partial, but I think this way of including it makes the most sense.

PR #568, opened 2017-07-28
[Commerce] Add hover states

raisaveuc:
# Add hover states:
* Inverted colors with easing for buttons (including landing page with
some custom responsive styles)
* Outline around amp-selector options
* Animate in underline for sidebar nav links, remove default underline
state
* Normal underline for misc text links
* Dark color and cursor: pointer for selector
* Underline radio options
* Underline product name for hover on related products and product listings

Other:
* Resolve workaround for
https://github.com/ampproject/amphtml/issues/10664


css/templates/e-commerce/page.css
```diff
@@ -701,6 +747,9 @@ amp-selector [option][selected] {
   border-bottom: 1px dashed #dadada;
 }
 
+.amp-mode-mouse .commerce-listing-product:hover > .commerce-listing-product-name {
```
camelburrito: i don't think hover works on mobile devices, so you can drop the amp-mode-mouse.  (same above)

raisaveuc: Some mobile devices still trigger the hover state on tap and remain until the user taps or navigates elsewhere. For this reason we only want to target devices that support a mouse pointer.

See first "Note" in docs for :hover
https://developer.mozilla.org/en-US/docs/Web/CSS/:hover

camelburrito: Can you add a comment explaining that?

raisaveuc: Added to the first instance in line 142

PR #571, opened 2017-08-01
[Commerce] Add focus to navbar trigger, update copy

raisaveuc:
* Adds back focus style for accessibility
* Moves custom CSS to JSON
* Updates placeholder text


templates/e-commerce/api/high-low-all-products.json
```diff
@@ -1,88 +1,88 @@
 {
```
raisaveuc: FYI all the json files in this folder are generated by the build.js script. They've been updated to have more varying placeholder text.

css/templates/e-commerce/page.css
```diff
@@ -330,12 +330,15 @@ amp-selector [option][selected] {
...
 }
 
-.ampstart-navbar-trigger {
-  position: absolute;
-  top: 0;
-  padding-top: var(--space-2);
+.ampstart-headerbar .ampstart-navbar-trigger {
   color: var(--color-primary);
   font-size: var(--h4);
+  /* Overrides pr2 default in navbar */
```
raisaveuc: Note that this is done here rather than the (navbar template)[https://github.com/raisaveuc/ampstart/blob/e20ca6709eed5a6c93f86e8d593106d5d30178c0/components/navbar/navbar.snip.html#L22] to minimize the impact on other pages that use this component

templates/e-commerce/api/high-low-all-products.json
camelburrito: Can you make the script - output that as a comment on the first line of this file.

raisaveuc: Comments are not supported in JSON. I tried different forms of JS comments anyway but it broke the page. I've added a file overview to build.js instead.

I've read suggestions on adding a __comments field to the beginning of the object, but I don't think this is an elegant solution. Did you have something else in mind or can we stick to the build.js comment?

PR #578, opened 2017-08-03
[Commerce] Add custom img preloader

raisaveuc:


css/templates/e-commerce/page.css
```diff
@@ -125,6 +125,62 @@ dd:after {
...
+
+.commerce-loader::before,
+.commerce-loader::after {
+  content: '';
+  position: absolute;
+  top: 0;
+}
+
+.commerce-loader::before {
+  left: -1rem;
```
raisaveuc: Using -1rem explicitly here and in keyframes instead of calc(-1 * var(--space-2)) to optimize performance/prevent the browser from calculating.

camelburrito:
Your templates have >40kb of css may be it is time to do some shaking up. (Split page level css if there are page level css components) @gurmukhp

gurmukhp:
Would it be worth adding uncss? I think most of our bloat comes from the BassCSS modules and inclining SVGs into the CSS.

PR #581, opened 2017-08-07
Built the webpack config for the AMP Start Configurator

torch2424:
This creates the web pack configuration for the AMP Start configurator. This supports the enviornments `dev` and `prod`. This also introduces some linting to the project, as it is used for the webpack process to stop unlinted builds from running. Also, confirmed that files are building to the correct directories, and dist files for configurator are located under `dist/configurator`

# Screenshots

**Please note:**
I have `webpack` alaised to `./node_modules/.bin/webpack --config webpack.config.js`

`dev`

screen shot 2017-08-07 at 3 11 15 pm

`prod`

screen shot 2017-08-07 at 3 11 40 pm

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -20,6 +20,7 @@ const ENV_ALIAS = {
     'prod'
   ],
   DEV: [
+    'development',
     'dev'
   ],
   TEST: [
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -51,7 +51,6 @@ function isNotEnv(envAlias) {
 module.exports = function (webpackEnv) {
   // Set our environemnt
   env = webpackEnv;
-  console.log(env);
   // Create our base webpack configuration
   const webpackConf = {
     module: {
@@ -94,10 +93,11 @@ module.exports = function (webpackEnv) {
   } else if (isEnv(ENV_ALIAS.DEV)) {
     webpackConf.module.loaders.push({
       test: /\.css$/,
-      loaders: ExtractTextPlugin.extract({
-        fallback: 'style-loader',
-        use: 'css-loader?minimize!postcss-loader'
-      })
+      loaders: [
+        'style-loader',
+        'css-loader',
+        'postcss-loader'
+      ]
     });
   }
 
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -106,6 +106,12 @@ module.exports = function (webpackEnv) {
     // Devtool
     webpackConf.devtool = 'source-map';
 
+    // Output
+    webpackConf.output = {
+      path: path.join(process.cwd(), conf.dest.configurator_tmp),
+      filename: 'index.js'
+    };
+
     // Entry
     webpackConf.entry = [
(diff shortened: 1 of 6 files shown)


webpack.config.js
```diff
@@ -0,0 +1,196 @@
...
+* Function to return if the environment matched the environment alias
+*/
+function isEnv(envAlias) {
+  // Check for default case
+  if (envAlias === ENV_ALIAS.DEV && (!env || Object.keys(env).length === 0)) {
+    return true;
+  }
+
+  return envAlias.some(alias => {
+    if (env[alias]) {
```
erwinmombay: can be shortened to `envAlias.some(alias => env[alias])`;

webpack.config.js
```diff
@@ -0,0 +1,196 @@
...
+        }
+      ]
+    },
+    plugins: [],
+    output: {},
+    entry: []
+  };
+
+  // LOADERS
+  if (isEnv(ENV_ALIAS.PROD)) {
```
erwinmombay: do you think we should do the multiple configs instead? the multiple ifs and different modifications to the config make it a little unreadable

webpack.config.js
```diff
@@ -0,0 +1,196 @@
...
+    'test'
+  ]
+};
+
+/**
+* Function to return if the environment matched the environment alias
+*/
+function isEnv(envAlias) {
+  // Check for default case
+  if (envAlias === ENV_ALIAS.DEV && (!env || Object.keys(env).length === 0)) {
```
erwinmombay: when is `env` falsy?

webpack.config.js
torch2424: I agree that it is a little harder to read. But having multiple files with slight changes is just as bad in my opinion. the `--env` is still relatively new, From what I can tell was introduced about a year ago, which is why you will not see a lot of webpack configs following the pattern because migrating can be a pain.

I'm open to either way at this point, but I personally feel a single file will be more maintainable

webpack.config.js
torch2424: env can be undefined if no environment is passed

webpack.config.js
erwinmombay: yes lets group by env, the multiple if conditions for the same environment make it hard to look at what an environments full configuration looks like

webpack.config.js
erwinmombay: but env is a module level variable that has a default empty object on line 14 right?

torch2424: Ahhhh You are correct, thank you! I'll fix this, my apologies

webpack.config.js
torch2424: Awesome, I'll get this done in my next round of changes

webpack.config.js
torch2424: done

webpack.config.js
torch2424: Done

webpack.config.js
torch2424: Done

PR #585, opened 2017-08-10
Amp Start Configurator Gulp Tasks

torch2424:
This PR is really on **118 additions and 15 deletions**, the `testsTemplates/article.html` can be ignored, and was approved by @camelburrito for testing.

This finishes up the build system for the amp start configurator. This creates all of the gulp tasks requried to build and serve the configurator. And I made it part of the standard build process. Also made gulp tasks for testing, and things of that sort.

# Example

**This example is just me serving the web application itself. On `npm run serve` the web app is watched, and will rebuild prod builds of the configurator. Which is what we want when developing both amp start and the configurator**

`npm run configurator` or `gulp configurator:serve`

configuratorampstart

### DIFF
--- a/tasks/configurator.js
+++ b/tasks/configurator.js
@@ -18,6 +18,7 @@
 const gulp = require('gulp-help')(require('gulp'));
 const config = require('./config');
 const util = require('gulp-util');
+const runSequence = require('run-sequence');
 const postcss = require('gulp-postcss');
 const replace = require('gulp-replace');
 const csstree = require('css-tree');
@@ -34,9 +35,24 @@ const webpackTestConfig = require('../webpack.config.js')({
   test: true
 });
 const WebpackDevServer = require("webpack-dev-server");
+const KarmaServer = require('karma').Server;
 
-function configuratorBuild() {
+function configuratorBuild(callback) {
+  runSequence('configurator:css', 'configurator:json', function(callback) {
+    webpack(webpackProdConfig, function(err, stats) {
+      if (err) {
+        throw new util.PluginError("webpack", err);
+      }
 
+      // Output the webpack stats
+      util.log("[webpack]", stats.toString({}));
+    });
+
+    // Call our passed callback
+    if(callback) {
+      callback();
+    }
+  });
 }
 
 function configuratorWatch() {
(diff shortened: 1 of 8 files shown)


tasks/configurator.js
```diff
@@ -13,15 +15,96 @@
...
+const webpackProdConfig = require('../webpack.config.js')({
+  prod: true
+});
+const webpackDevConfig = require('../webpack.config.js')({
+  dev: true
+});
+const webpackTestConfig = require('../webpack.config.js')({
+  test: true
+});
+const WebpackDevServer = require("webpack-dev-server");
```
erwinmombay: use single quote here and anywhere else possible

tasks/configurator.js
```diff
@@ -13,15 +15,96 @@
...
+    }
+
+    // Output the webpack stats
+    util.log("[webpack]", stats.toString({}));
+  });
+}
+
+function runWebpackDevServer(webpackConfig) {
+  if(webpackConfig.entry && Array.isArray(webpackConfig.entry)) {
+    webpackConfig.entry.unshift("webpack-dev-server/client?http://localhost:8080");
```
erwinmombay: lets make the port configurable with the default as 8080

tasks/configurator.js
```diff
@@ -13,15 +15,96 @@
...
+
+function runKarma(options, callback) {
+
+  // Parse our karma config
+  karmaServerConfig = {
+    configFile: `${__dirname}/../karma.conf.js`
+  };
+
+  // Pass in our options
+  Object.keys(options).forEach(function(key) {
```
erwinmombay: We can use Object.assign here i think to make this shorter (depending on support)

torch2424: Hahahahaha! You are right, that is what I get for copying code from the webpack docs. Thanks!

tasks/configurator.js
```diff
@@ -13,15 +15,96 @@
...
 const postcss = require('gulp-postcss');
 const replace = require('gulp-replace');
 const csstree = require('css-tree');
 const rename = require('gulp-rename');
 const through = require('through2');
+const webpack = require("webpack");
+const webpackProdConfig = require('../webpack.config.js')({
+  prod: true
+});
+const webpackDevConfig = require('../webpack.config.js')({
```
erwinmombay: love this! +1

tasks/configurator.js
torch2424: Done

tasks/configurator.js
torch2424: Done

tasks/configurator.js
torch2424: Done

tasks/configurator.js
torch2424: Thanks!

PR #593, opened 2017-08-17
[Configurator] Added Web Workers to Support CSS Transpilation

torch2424:
* Offloaded postcss transpilation into a webworker, and updated the `CssTranspiler.getCssWithVars()` to return a promise accordingly.

Also, I profiled switching variables very fast and quickly, and handled the operation fine. No more than 10% CPU usage, and ~170MB of Ram under load, ~9MB idle (For comparison, Gmail uses ~130MB idle).

# Screenshot

screen shot 2017-08-16 at 5 16 23 pm

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -95,7 +95,13 @@ module.exports = function (webpackEnv) {
         new webpack.HotModuleReplacementPlugin(),
         new webpack.LoaderOptionsPlugin({
           options: {
-            postcss: () => [autoprefixer]
+            postcss: () => [autoprefixer],
+            worker: {
+              output: {
+                filename: 'hash.worker.js',
+                chunkFilename: '[id].hash.worker.js'
+              }
+            }
           },
           debug: true
         }),
@@ -153,7 +159,13 @@ module.exports = function (webpackEnv) {
         new webpack.optimize.CommonsChunkPlugin({name: 'vendor'}),
         new webpack.LoaderOptionsPlugin({
           options: {
-            postcss: () => [autoprefixer]
+            postcss: () => [autoprefixer],
+            worker: {
+              output: {
+                filename: 'hash.worker.js',
+                chunkFilename: '[id].hash.worker.js'
+              }
+            }
           }
         })
       ]);
@@ -180,7 +192,14 @@ module.exports = function (webpackEnv) {
     webpackConf.plugins =
       webpackConf.plugins.concat([
         new webpack.LoaderOptionsPlugin({
-          options: {},
+          options: {
+            worker: {
+              output: {
+                filename: 'hash.worker.js',
+                chunkFilename: '[id].hash.worker.js'
+              }
+            }
+          },
           debug: true
         })
       ]);
--- a/www/configurator/src/app/css-transpiler/css-transpiler.js
+++ b/www/configurator/src/app/css-transpiler/css-transpiler.js
@@ -21,6 +21,10 @@
 
 const postcss = require('postcss');
 const customProperties = require('postcss-custom-properties');
+// Grab the Worker Using Webpack
+// https://github.com/webpack/webpack/tree/master/examples/web-worker
+// https://github.com/webpack-contrib/worker-loader
+const Worker = require('worker-loader!./css-transpiler.worker');
 
 class CssTranspiler {
   /**
@@ -39,6 +43,11 @@ class CssTranspiler {
    * @returns {string} - the newly transpiled page css with varibale values
    */
   getCssWithVars(passedCssVars) {
+    // Testing the webworker
+    // eslint-disable-next-line
+    const worker = new Worker;
+    worker.postMessage('b');
+
     // Only assign variables that exist in both, and set to the current value
     const cssVars = Object.assign({}, this.templateCssVars);
     Object.keys(passedCssVars).forEach(cssVarKey => {
@@ -46,6 +55,8 @@ class CssTranspiler {
         cssVars[cssVarKey].current = passedCssVars[cssVarKey];
       }
     });
+
+    // Add the onmessage here
 
     // Append the new vars to the end of our template css
     let cssWithAppendedVars = `${this.templateCss.slice(0)} :root {`;
--- a/www/configurator/src/app/css-transpiler/css-transpiler.js
+++ b/www/configurator/src/app/css-transpiler/css-transpiler.js
@@ -32,8 +32,12 @@ class CssTranspiler {
   constructor(css, cssVars) {
     this.templateCss = css;
     this.templateCssVars = cssVars;
-    // eslint-disable-next-line
-    this.worker = new Worker;
+    /* eslint-disable */
+    this.asyncPostcssWorker = {
+      worker: new Worker,
+      resolved: true
+    }
+    /* eslint-enable */
   }
 
   /**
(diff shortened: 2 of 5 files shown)


www/configurator/src/app/css-transpiler/css-transpiler.js
```diff
@@ -30,15 +32,34 @@ class CssTranspiler {
   constructor(css, cssVars) {
     this.templateCss = css;
     this.templateCssVars = cssVars;
+    /* eslint-disable */
```
torch2424: I originally did this to avoid the http://eslint.org/docs/rules/new-parens error. Since the webpack docs didn't use the () when creating new Workers. But I have noticed it does not make a different, so I will go ahead and remove this

done

www/configurator/src/app/css-transpiler/css-transpiler.worker.js
```diff
@@ -0,0 +1,27 @@
...
+	* 		Simply responds to messages containing the template CSS, and css vars, and responds with the newly transpiled css.
+  */
+
+const postcss = require('postcss');
+const customProperties = require('postcss-custom-properties');
+
+onmessage = function(event) {
+
+	// Obtain the templateCss and cssvars from event
+	const templateCss = event.data.templateCss;
```
torch2424: I agree.

Edge cases I can check for are for missing data, and invalid cssVars.

Any others?

www/configurator/src/app/css-transpiler/css-transpiler.worker.js
```diff
@@ -0,0 +1,27 @@
...
+	let cssWithAppendedVars = `${templateCss} \n:root {`;
+	Object.keys(cssVars).forEach(cssVarKey => {
+		cssWithAppendedVars += `\n${cssVarKey}: ${cssVars[cssVarKey].current || cssVars[cssVarKey].value};`;
+	});
+	cssWithAppendedVars += '\n}';
+
+	// Transpile the CSS with the appended variables
+	postMessage(postcss().use(customProperties()).process(cssWithAppendedVars).css);
+}
+/* eslint-enable */
```
torch2424: I agree, I found a nice way to simply satisfy the original linting issue of having the onmessage function being defined inside the worker.

done

www/configurator/src/app/css-transpiler/css-transpiler.worker.js
camelburrito: anything you can think of - obvious ones are missing data - invalid vars might be a good one - we might need to let the users know what went wrong

www/configurator/src/app/css-transpiler/css-transpiler.worker.js
```diff
@@ -0,0 +1,27 @@
...
+
+	// Append the new vars to the end of our template css
+	let cssWithAppendedVars = `${templateCss} \n:root {`;
+	Object.keys(cssVars).forEach(cssVarKey => {
+		cssWithAppendedVars += `\n${cssVarKey}: ${cssVars[cssVarKey].current || cssVars[cssVarKey].value};`;
+	});
+	cssWithAppendedVars += '\n}';
+
+	// Transpile the CSS with the appended variables
+	postMessage(postcss().use(customProperties()).process(cssWithAppendedVars).css);
```
torch2424: After looking through the docs, this actually does not throw any errors.

However, there is a promise syntax that is suggested over this one, incase we decide to add any asynchronous plugins to our worker. So for that case, I'll go ahead and do that.

Here is the API Docs on what process() returns: http://api.postcss.org/LazyResult.html#then

www/configurator/src/app/css-transpiler/css-transpiler.js
```diff
@@ -47,15 +68,18 @@ class CssTranspiler {
       }
     });
 
-    // Append the new vars to the end of our template css
-    let cssWithAppendedVars = `${this.templateCss.slice(0)} :root {`;
-    Object.keys(cssVars).forEach(cssVarKey => {
-      cssWithAppendedVars += `\n${cssVarKey}: ${cssVars[cssVarKey].current || cssVars[cssVarKey].value};`;
+    // Add the onmessage here
+    this.asyncPostcssWorker.worker.postMessage({
+      templateCss: this.templateCss.slice(0),
```
torch2424: Originally I was worred about the original reference being modified, but it is not after I made some changes.

done

webpack.config.js
```diff
@@ -153,7 +159,13 @@ module.exports = function (webpackEnv) {
         new webpack.optimize.CommonsChunkPlugin({name: 'vendor'}),
         new webpack.LoaderOptionsPlugin({
           options: {
-            postcss: () => [autoprefixer]
+            postcss: () => [autoprefixer],
+            worker: {
+              output: {
+                filename: 'hash.worker.js',
```
torch2424: These lines allow for require statements in the form of: 

```
const Worker = require('worker-loader![RELATIVE FILE PATH HERE]');
```

To be loaded into their own worker.js file, which a hash id for cache busting on the client.

This is required since [Web workers must be created from a URI](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers#Spawning_a_dedicated_worker) And cannot be "officially" loaded in the same JS file. [There is a hack for this](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers#Embedded_workers), but I would prefer to have webpack handle this, than us try to inline the JS ourselves into a script tag

www/configurator/src/app/css-transpiler/css-transpiler.worker.js
camelburrito: may be we should move it in a try/catch?

torch2424: Done

webpack.config.js
torch2424: Done

www/configurator/src/app/css-transpiler/css-transpiler.worker.js
torch2424: Yes, I agree, lets talk offline about that tommorow :)

webpack.config.js
camelburrito: isn't our webworker in its own JS file?

www/configurator/src/app/css-transpiler/css-transpiler.js
```diff
@@ -30,15 +32,30 @@ class CssTranspiler {
...
-   * @returns {string} - the newly transpiled page css with varibale values
+   * @returns {Promise} - Will resolve with the response from a webworker, containing the newly transpiled css
    */
   getCssWithVars(passedCssVars) {
+    // First check if we have a worker, and if it didRespond
+    if (this.asyncPostcssWorker.worker && !this.asyncPostcssWorker.didRespond) {
+      // Terminate the worker, and recreate it
+      this.asyncPostcssWorker.worker.terminate();
+      this.asyncPostcssWorker = {
+        worker: new Worker(),
```
camelburrito: where is the worker filename being passed?

torch2424: The filename is being passed on about line 25, when we Define/require the Worker.

The `worker-loader!` tells the web pack to use the worker-loader, and to use the filepath that is following it.

camelburrito: can we rename Worker to CSSTranspilerWroker

torch2424: Of course 😄

torch2424: Done

www/configurator/src/index.js
```diff
@@ -77,18 +77,24 @@ Promise.all(configuratorInit).then(responses => {
...
  */
 function handleHashChange_() {
-  const updatedStyles = cssTranspiler.getCssWithVars(getUrlCssVars());
-  iframeManager.setStyle(updatedStyles);
+  cssTranspiler.getCssWithVars(getUrlCssVars()).then(updatedStyles => {
+    iframeManager.setStyle(updatedStyles);
+  }).catch(() => {
+    // Set styles back to the initial styles
+    // TODO (torch2424 && camelburrito) - Talk offline to decide how to display transpilation error to user
+    iframeManager.setStyle(cssTranspiler.templateCss);
```
camelburrito: this should be previously used styles (or don't touch the iframe.)

torch2424: So would you like me to just do nothing in the catch block? Since the styles aren't applied unless the worker resolves successfully

torch2424: Talked offline, a UI for error messages will be dealt with later. Simply bubble the error to here

PR #613, opened 2017-09-15
[LAND-SEE] - Add instagram CTA, accordion sections, stub api data

afilbert:
PR covers `amp-anim` responsive instagram call to action, and accordion sections loaded via `amp-list` and stub api endpoints for posts. Posts are animated to slide and fade in.

Firebase staging deployed:

Please note that this isn't validating because of the `amp-animation` scripts.


css/templates/land-see/page.css
```diff
@@ -354,3 +354,115 @@ amp-img.cover img {
 .ampstart-input input[type="email"].user-invalid {
   border: 2px solid var(--color-form-invalid);
 }
+
+/* Accordion */
```
camelburrito: can you give this a meaningful name (land-see class). Changing all the accordions is not a good idea, imagine if we add a component later that uses accordion then this template would be broken.

css/templates/land-see/page.css
```diff
@@ -354,3 +354,115 @@ amp-img.cover img {
...
+}
+
+.i-amphtml-accordion-header { /* reset defaults */
+  cursor: default;
+  background-color: inherit;
+  padding-right: 0;
+  border: none;
+}
+
+section.accordion-bg-100:before {
```
camelburrito: again `accordion-bg-100` is not a good class name - ideally namespace your classes and name based on functionality not utility  (the only utility classes we allow are from BASS CSS.)

templates/land-see/land-see.amp.html
```diff
@@ -42,10 +61,82 @@ <h1 class="pt3">Hello AMP!</h1>
...
+        Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo. Nemo enim ipsam voluptatem quia voluptas sit aspernatur aut odit aut fugit, sed quia consequuntur magni dolores eos qui ratione voluptatem sequi nesciunt.
       </p>
     </div>
 
+    {{#accordion}}
+    <div id="accordion-container" class="accordion center items-center">
```
camelburrito: I think this should be a `<section>` give the section a namespaced class name , I see that all your names are missing the page(land-see) namespace

templates/land-see/land-see.amp.html
```diff
@@ -42,10 +61,82 @@ <h1 class="pt3">Hello AMP!</h1>
...
+        Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo. Nemo enim ipsam voluptatem quia voluptas sit aspernatur aut odit aut fugit, sed quia consequuntur magni dolores eos qui ratione voluptatem sequi nesciunt.
       </p>
     </div>
 
+    {{#accordion}}
```
camelburrito: i am okay moving this into a partial if you think it is required.

templates/land-see/land-see.amp.html
```diff
@@ -42,10 +61,82 @@ <h1 class="pt3">Hello AMP!</h1>
...
+    <div id="accordion-container" class="accordion center items-center">
+      <amp-accordion layout="container" disable-session-states class="amp-accordion">
```
camelburrito: amp-accordion is not a good class name -

Here is what i think i would do (i am not 100% sure if what name would be  appropriate)

```html
<section class="land-see-stories">
    <amp-accordion> <!-- Now that you have a wrapping name space you can target using that class or add anothet class here -->
</section>
```



sidenote: Take a look the guidelines doc Eric had sent you.

templates/land-see/land-see.amp.html
```diff
@@ -42,10 +61,82 @@ <h1 class="pt3">Hello AMP!</h1>
...
+            </amp-list>
+          </div>
+        </section>
+        {{/section}}
+      </amp-accordion>
+    </div>
+    {{/accordion}}
+
+    {{#instagram-cta}}
+    <div class="cta-wrapper mx-auto">
```
camelburrito: why do we need this wrapper? can't we make the a block (and give it a nice class name to set the max-width 
 or use bass - max-width-1/max-width-2 etc)and mx-auto?

templates/land-see/land-see.amp.html
afilbert: I didn't realize that the CSS written for within the land-see template would affect the broader project, @camelburrito. Thanks for clarifying. Will clean this and other references up in my next commit. :)

templates/land-see/land-see.amp.html
afilbert: The accordion only appears on the "home" template page in this project, which is why I didn't move it into a partial. Please let me know if I should do that in the interest of keeping the markup cleaner (regardless of reuse).

css/templates/land-see/page.css
afilbert: Yes, good catch. I name-spaced the other form email input, but not the error handling.

templates/land-see/land-see.amp.html
afilbert: Will do.

templates/land-see/land-see.amp.html
afilbert: The template calls for different y-margins than I had available in Basecss, at different breakpoints. I created this wrapper to follow that behavior on top of the media calls I already was passing into the `amp-anim` components. Let me play with bass and see about simplifying.

afilbert:
Thanks for the feedback, @camelburrito! I'll definitely be sure to namespace my classes better going forward with my next commit and subsequent PRs.

afilbert:
@camelburrito went ahead and pushed those changes. Please let me know if you need anything else.

css/templates/land-see/page-vars.css
```diff
@@ -60,6 +60,12 @@
   --color-input-secondary: var(--color-font-secondary);
   --color-input-font-secondary: var(--color-font-secondary);
 
+  --color-accordion-color-base: #FDDCD7;
```
camelburrito: rename this variable,  also no caps in color values :)

css/templates/land-see/page-vars.css
```diff
@@ -60,6 +60,12 @@
...
+  --opacity-100: 1;
```
camelburrito: I dont think these need to be variables (also dont create bass style variables, it would be confusing when we build a configurator.)

css/templates/land-see/page.css
```diff
@@ -351,6 +350,119 @@ amp-img.cover img {
...
+}
+
+/* Post */
+
+.land-see-post-item {
+  max-width: 196px;
+  text-align: left;
+}
+
+.land-see .animate {
```
camelburrito: animate is not a good class name (namespace + functionality)

css/templates/land-see/page.css
```diff
@@ -351,6 +350,119 @@ amp-img.cover img {
...
+  opacity: 0;
+  transform: translateY(-10px);
+}
+
+.land-see-post-item.wide {
```
camelburrito: wide is not a good class name 

you can either have 

`land-see-post-item-wide` or use media query (may be mediaquery is the riht thing here)

css/templates/land-see/page.css
```diff
@@ -351,6 +350,119 @@ amp-img.cover img {
...
+  opacity: var(--opacity-77);
+}
+
+.land-see-post-item p {
+  font: 400 var(--h6)/var(--line-height-2) var(--font-primary);
+}
+
+/* Instagram/Sale CTAs */
+
+.land-see-cta-link {
```
camelburrito: i dont know if the name has to be generic (you take the call). To me CTA sounds very generic , the comment above says its for sale.

templates/land-see/land-see.amp.html
camelburrito: I think it might be cleaner in a partial (you can make the call), this works too.

templates/land-see/land-see.amp.html
```diff
@@ -42,10 +61,80 @@ <h1 class="pt3">Hello AMP!</h1>
...
+        Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo. Nemo enim ipsam voluptatem quia voluptas sit aspernatur aut odit aut fugit, sed quia consequuntur magni dolores eos qui ratione voluptatem sequi nesciunt.
       </p>
     </div>
 
+    {{#accordion}}
```
camelburrito: can you rename the json var too

templates/land-see/land-see.amp.html
```diff
@@ -42,10 +61,80 @@ <h1 class="pt3">Hello AMP!</h1>
...
+    <section class="land-see-accordion center items-center">
+      <amp-accordion layout="container" disable-session-states>
+        {{#section}}
+        <section {{expanded}} class="{{classes}}">
+          <h3 class="max-width-4 mx-auto">{{title}}</h3>
+          <div class="accordion-content max-width-4 mx-auto">
```
camelburrito: namespaced class names

templates/land-see/land-see.amp.html
```diff
@@ -42,10 +61,80 @@ <h1 class="pt3">Hello AMP!</h1>
...
+        Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo. Nemo enim ipsam voluptatem quia voluptas sit aspernatur aut odit aut fugit, sed quia consequuntur magni dolores eos qui ratione voluptatem sequi nesciunt.
       </p>
     </div>
 
+    {{#accordion}}
+    <section class="land-see-accordion center items-center">
```
camelburrito: `land-see-accordion` is names-paced , but it still does not say what it is for. To some one trying to use this it would feel like this applies to all accordions in this template. If you give the right name (somehow conveying what this section is for) that would be useful

templates/land-see/land-see.amp.html
afilbert: Another good catch. Both that class and amp-accordion came from `amp-accordion` example code, and are extraneous. Will remove. I'll watch out for those in future PRs.

css/templates/land-see/page-vars.css
afilbert: Makes sense.

css/templates/land-see/page.css
afilbert: This isn't used until a future PR, but will revise CSS to reflect your comments.

afilbert: For added context on some of my CSS class naming... I got ahead of the review game before our back and forth on the last few PRs, and noticed the growing CSS file size for land-see as I added features. So I actually backtracked on some of the CSS namespacing in my local and fork branches, and tried to keep class names short and sweet for that reason. My goal was to make more room for CSS additions. We have about 15KB before we hit our 50K limit.

css/templates/land-see/page.css
afilbert: This naming, as with the accordion, came directly from the design commentary in Basecamp. There are two CTAs (calls to action) on the homepage template:

* Summer Sale (a future PR)
* Instagram

They're very similar, so I wanted to use the same classes for both.

templates/land-see/land-see.amp.html
afilbert: As mentioned in my other [comment](https://github.com/ampproject/ampstart/pull/613#discussion_r139295000), I've been following the design process language. I'll add some context around the story categories.

afilbert:
Thanks for the added direction, @camelburrito. I think this one's ready to go!

camelburrito:
https://travis-ci.org/ampproject/ampstart/builds/276349577?utm_source=github_status&utm_medium=notification - can you check why this is failing  

@afilbert

css/templates/land-see/page.css
```diff
@@ -351,6 +350,119 @@ amp-img.cover img {
...
+.land-see-story-category section:before {
+  content: '';
+  position: absolute;
+  height: 168px;
+  left:0;
+  right: 0;
+  background-color: var(--color-story-category-background);
+}
+
+.land-see .i-amphtml-accordion-header { /* reset defaults */
```
camelburrito: aha - this is invalid amp :) 

If this applies to all the accordions on the page 
just do `amp-accodion > section > header` , it would be easy for you if you assign a class to the accordions. remember sometimes we might use accodions to do drop down menus in the AMP Start components. The ideal way to solve this is go give a name like .land-see-accordion (if this applies to all page specific accordion)

css/templates/land-see/page.css
```diff
@@ -351,6 +350,119 @@ amp-img.cover img {
...
+  cursor: default;
+  background-color: inherit;
+  padding-right: 0;
+  border: none;
+}
+
+.land-see-story-category section.interiors:before {
```
camelburrito: land-see-interiors

I guess you could come up with an additional class name on the category 

`land-see-story-interior` now you just need to use only one class to target

Sometimes adding extra classes would save you size

camelburrito:
I know you mentioned about increasing size , we usually counter that my maxing out on basscss classes, targeting using one class. 

Your template is an opinionated template so it is okay to go up on size till you think it is comfortable.

css/templates/land-see/page.css
afilbert: Good suggestion, that'll help trim some CSS fat.

afilbert:
Good to know on the CSS size limitation. I'll also keep your better targeting direction in mind. Was able to apply that pretty broadly to the category accordion work. 

Thanks for discovering that invalid CSS. I was missing that among the disallowed script validation errors. I've pushed the latest up, @camelburrito.

PR #615, opened 2017-09-18
[LAND-SEE] - Add summer sale CTA, background image animation

afilbert:
PR adds Summer Sale section (yay for alliteration). I use that to anchor and position the background triangles, and animate them on scroll.

Safari presented significant challenges:

* Right background image (pink-colored downward triangle) is causing horizontal scrolling.
* Either `amp-animation` or `amp-position-observer` is failing to animate on scroll. Issue, [here](https://github.com/ampproject/amphtml/issues/11311).

I tried a number of approaches to eliminate the horizontal scroll in Safari, including positioning using negative margins, background images through pseudo-classes, absolute positioning, and nested wrapper elements, but it seems to come down to some amphtml boilerplate:

```
html.i-amphtml-singledoc > body {
    overflow: visible !important;
    position: relative !important;
}
```

Changing `overflow: visible !important` to `overflow: hidden !important` in the inspector fixes the issue on Safari. I can't override because of the `!important` tag, however, and it looks like that's by design.

Staging URL: https://landseeampstart.firebaseapp.com/templates/land-see/land-see.amp


spacedino:
@camelburrito is this fixed yet?

PR #625, opened 2017-09-23
[LAND-SEE] - Add popular with responsive carousel, recent with accent animation

afilbert:
PR adds popular story posts and recent sections. Staging URL:

Of note is the structure of the popular.json data, which I used to dynamically generate multiple `amp-carousel` components; each with their own media behaviors. That required repeating the data depending on the slide layout at different responsive breakpoints.

Another highlight is that CSS_SIZES seemed to slightly change for each page between my last PR and this. That's likely going to cause some rebase conflicts for others.


css/templates/land-see/page.css
```diff
@@ -49,6 +49,13 @@ h4,
   line-height: var(--line-height-2);
 }
 
+h4.land-see-section-header {
```
camelburrito: i think you can drop the h4 (is it needed?)

css/templates/land-see/page.css
```diff
@@ -514,3 +532,35 @@ h4,
...
+
+/* Popular/Recent/Reccomended Stories */
+
+.land-see-popular-content,
+.land-see-recommended-content,
+.land-see-recent-content {
+  max-width: 850px;
+}
+
+.land-see-popular-content .amp-carousel-button,
```
camelburrito: don't you want all the carousels in this theme/template to have these arrows? Remember some one who downloads this template could also be using  something from ampstart.com/components#media. I think you can override all the arrows - what do you think!

afilbert: I do want the arrows to come from the same background SVG CSS for all carousels in the template, yes. The hero carousel arrows are bigger than the popular and recommended ones, however, so I was planning on applying a separate style to those and overriding the `background-size` property. I'm cool with simplifying the CSS here, though, too.

afilbert:
Thanks for the feedback, @camelburrito! Please take a look at my latest commit and let me know if you need anything else.

PR #627, opened 2017-10-01
[LAND-SEE] - Add responsive hero carousel

afilbert:
PR adds responsive hero carousel, and wraps up work on the home template. Staging URL:

Similar to #625, carousel data is repeated several times in `land-see.amp.json` in order to achieve the desired responsiveness and behavioral results.


css/templates/land-see/page.css
```diff
@@ -566,3 +567,156 @@ h4,
...
+}
+
+.land-see-hero-typography {
+  bottom: -7%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+}
+
+.land-see-hero-typography a {
+  text-decoration: none;
```
camelburrito: you can use the basscss classes for this .

css/templates/land-see/page.css
```diff
@@ -566,3 +567,156 @@ h4,
...
+}
+
+.land-see-hero-title {
+  font: 400 var(--h1)/var(--line-height-6) var(--font-primary);
+  letter-spacing: normal;
+}
+
+.land-see-hero-caption {
+  width: 28rem;
+  font: 400 var(--h7)/var(--line-height-6) var(--font-primary);
```
camelburrito: i think for a lot of these you would be fine overriding variables or rewriting the typography css classes (not sure if you had already explored that)

afilbert:
Cleaned up the CSS, @camelburrito. Was able to find some good uses for basecss and base-override-vars as well. Please let me know if you need anything else.

PR #631, opened 2017-10-05
[GALLERY] Add sidebar component

bentranter:
Adds the sidebar component (along with its required header component), which is responsible for expanding and collapsing the sidebar.

Please note that the images used here are the same as the ones that will be used in the next PR, hence the large image size.

cc @camelburrito 

EDIT: Link to staging site for this PR http://ampstart-pr-gallery-sidebar-2vwqha3eha1.surge.sh/templates/gallery/gallery.amp (it's just the sidebar for now, click the hamburger icon to see that sidebar content)

PR #646, opened 2017-10-11
[GALLERY] WIP - Sidebar component

bentranter:
As mentioned in #631, this is a WIP PR for a Sidebar that reuses an Ampstart component - the `navbar`.

I'm not sure if this approach is appropriate, so I was hoping to get some insight from @camelburrito.

There are two main issues I see with it:

1. Using an `::after` pseudo element's `content` rule to show the "Current Exhibitions" text, as it's not properly screen-reader friendly (as far as I know).
2. The only way to get images and SVG icons into the sidebar `links` is to insert them as HTML in the `gallery.amp.json` file. This doesn't feel right to me, but I'm curious what you think about it.

Anyway, let me know if this approach is ok, or if there's something different we should we do.

**Note: It's probably most efficient** to just to check the `templates/gallery/gallery.amp.html` file and the `templates/gallery/gallery.amp.json` file for now instead of reviewing the entire PR, since those are the ones in question.

We're also just waiting on the upload to get a staging branch ready to look at, so I'll edit this once I've got it.

[EDIT]: Staging link for this PR: https://ampstart-staging-metalab-f14d646917de.surge.sh/templates/gallery/gallery.amp.html


camelburrito:
@bentranter how is this different from your previous PR? I dont  see sidebar being reused from components/

bentranter:
@camelburrito ah sorry, I forgot to link the other file - we reuse the `navbar` component here [#diff-f2f757e5](https://github.com/ampproject/ampstart/pull/646/files#diff-f2f757e5b540207624fd5e702e592940R20), and populate it with content here [#diff-146c665d](https://github.com/ampproject/ampstart/pull/646/files#diff-146c665d0dd94d6062c61f1fb278d3d5R20).

camelburrito:
@bentranter so delete the header and sidebar snip files that you added?

10xjs:
@camelburrito We need your feedback on the current work in progress integrating the shared sidebar component. Ben has some concerns on some of the workarounds he as had to make - they might outweigh the benefits of using the shared component.

camelburrito:
@bentranter - i think you can add that text (passed in via json) to the sidebar - i guess we already have a name attribute that you can use (inside `{{logo}}`). 
feel free to create something like {{display-in-sidebar}} boolean in that and pass in the text that way. The components can be subject to modification in a way that is backward compatible

camelburrito:
2. Yes you can use SVG for now , we have had this problem for few other pages too. Ill work on figuring out an elegant solution and port yours.

bentranter:
@camelburrito Thanks for the suggestion and feedback! For now I'll go with using the name attribute inside `{{logo}}`. While I think a boolean in the sidebar itself would be the ideal solution, I'm worried about the time it'd take for me to verify that adding that boolean wouldn't break any existing uses of the `sidebar` component (especially with Lune's uses of it), since I'm not too familiar with the component outside my own use of it here. If that's something you want to create, I'd be happy to use it, but if not, I'll use the `{{logo}}` to keep this PR as simple as possible.

camelburrito:
>  I'm worried about the time it'd take for me to verify that adding that boolean wouldn't break any existing uses of the sidebar component (especially with Lune's uses of it), since I'm not too familiar with the component outside my own use of it here

As long as you add the new boolean that no one uses you are not going to break anything dont worry!

camelburrito:
@bentranter - just checking are you waiting on me or is this being worked upon?

bentranter:
@camelburrito it's being worked on :)

bentranter:
@camelburrito ready for re-review

templates/gallery/gallery.amp.json
```diff
@@ -16,7 +16,68 @@
...
+            "text": "<div class='gallery-sidebar-nav-item-text'>Short Circuit<br/>Anna Moss</div><amp-img class='gallery-sidebar-nav-item-image' width='1' height='1' layout='responsive' src='../../img/gallery/lamp.jpg' srcset='../../img/gallery/lamp.jpg 1x, ../../img/gallery/lamp@2x.jpg 2x'></amp-img><div class='gallery-sidebar-nav-item-icon'><svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100 100'><polyline fill='none' stroke='currentColor' points='27 5 72.167 50.167 27.335 95'/></svg></div>"
+          }
+        },
+        {
+          "link": {
+            "nav-item-classes": ["gallery-sidebar-nav-item"],
+            "url": "./gallery-inner.amp.html",
+            "text": "<div class='gallery-sidebar-nav-item-text'>Geo Breakfast<br/>Terra Morales</div><amp-img class='gallery-sidebar-nav-item-image' width='1' height='1' layout='responsive' src='../../img/gallery/balloon.jpg' srcset='../../img/gallery/balloon.jpg 1x, ../../img/gallery/balloon@2x.jpg 2x'></amp-img><div class='gallery-sidebar-nav-item-icon'><svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100 100'><polyline fill='none' stroke='currentColor' points='27 5 72.167 50.167 27.335 95'/></svg></div>"
```
camelburrito: could we not do the SVG as a background image on ::after?

bentranter: I'd mentioned this on Friday when we spoke, and have spent some time on it since then, but I wasn't able to get this to work.

bentranter:
@camelburrito were there any other changes that you wanted to this PR?

PR #666, opened 2017-10-13
[TRAVEL] Set up initial project structure

bentranter:
Adds the base templates, CSS, and JSON files for the upcoming Travel template.

cc @camelburrito

### DIFF
--- a/css/templates/travel/page-vars.css
+++ b/css/templates/travel/page-vars.css
@@ -16,6 +16,7 @@
 
 :root {
   --font-color-primary: var(--black);
+  --font-color-secondary: var(--white);
   --font-family-primary: "Source Sans Pro", -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", OpenSansEmoji, EmojiSymbols, EmojiSymb, Symbola;
   --line-height: 1.2;
   --fixed-navbar-height: calc(86 / 16rem);;
--- a/css/templates/travel/page-vars.css
+++ b/css/templates/travel/page-vars.css
@@ -16,10 +16,27 @@
 
 :root {
   --font-color-primary: var(--black);
+  --color-primary: var(--black);
   --font-color-secondary: var(--white);
+  --color-secondary: var(--white);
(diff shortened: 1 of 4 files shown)


css/templates/travel/page-vars.css
```diff
@@ -0,0 +1,96 @@
...
+  --type-fashion-color: #f1c363;
+  --type-food-color: #aa5efd;
+  --type-music-color: #ec8936;
+  --type-nature-color: #14b52a;
+  --type-nightlife-color: #e83a3a;
+  --type-tours-color: #0ece9e;
+  --type-water-color: #3abee8;
+
+  /* input element */
+  --input-padding-y: var(--button-padding-y);
```
camelburrito: do you need these? can't you use the `p*` classes

css/templates/travel/page-vars.css
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+:root {
+  --font-color-primary: var(--black);
```
camelburrito: you will need to set the secondary colors for this to work right

camelburrito:
@bentranter - 2 minor comments

css/templates/travel/page-vars.css
bentranter: I believe we need them, with BassCSS you can't use the `p*` classes to style the input element because it gets overwritten by BassCSS's `input[type=text]` rule, so this is used to override that.

css/templates/travel/page-vars.css
bentranter: 👍 Cool, I'll set it

camelburrito: also set color-primary and secondary

camelburrito: https://github.com/ampproject/ampstart/blob/master/css/ampstart-base/base-page-vars.css - Look at this and set anything that you should (setting all is great :))

bentranter: Everything that we can use should be set now

PR #673, opened 2017-10-17
[GALLERY] Add landing page with art display

bentranter:
Adds the core part of the landing page, which contains the art displays. Since this code is pretty well-separated from the sidebar, I figured it'd be worthwhile to get this PR up as well.

Staging link: https://ampstart-pr-gallery-art-display-asg3asdg38sb3y.surge.sh/templates/gallery/gallery.amp

cc @camelburrito


templates/gallery/partials/landing-images.snip.html
```diff
@@ -0,0 +1,42 @@
...
+-->
+}}
+
+{{#landing-images}}
+  <figure class="gallery-landing-image-wrap-{{index}}">
+    <a href="./gallery-inner.amp.html" class="gallery-landing-image-link"></a>
+    <figcaption class="gallery-landing-image-caption {{#positions}}gallery-landing-image-caption-{{.}} {{/positions}} gallery-landing-image-caption-content">
+      <div>{{title}}</div>
+      <div>{{artist}}</div>
+      <a href="./gallery-inner.amp.html" class="gallery-landing-caption-link inline-block">View Exhibit</a>
```
camelburrito: when will this be visible?

templates/gallery/partials/landing-images.snip.html
```diff
@@ -0,0 +1,42 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+}}
+
+{{#landing-images}}
+  <figure class="gallery-landing-image-wrap-{{index}}">
+    <a href="./gallery-inner.amp.html" class="gallery-landing-image-link"></a>
+    <figcaption class="gallery-landing-image-caption {{#positions}}gallery-landing-image-caption-{{.}} {{/positions}} gallery-landing-image-caption-content">
+      <div>{{title}}</div>
```
camelburrito: h* tag

templates/gallery/partials/landing-images.snip.html
```diff
@@ -0,0 +1,42 @@
...
+      <div>{{artist}}</div>
```
camelburrito: ideally i would use semantics similar to byline (look for the roles and aria)

templates/gallery/partials/landing-images.snip.html
```diff
@@ -0,0 +1,42 @@
...
+distributed under the License is distributed on an "AS-IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+}}
+
+{{#landing-images}}
+  <figure class="gallery-landing-image-wrap-{{index}}">
+    <a href="./gallery-inner.amp.html" class="gallery-landing-image-link"></a>
```
camelburrito: i think the anchor should wrap the entire figure.

templates/gallery/partials/landing-images.snip.html
```diff
@@ -0,0 +1,42 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS-IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+}}
+
+{{#landing-images}}
```
camelburrito: To me this looks exactly like "image-with-caption", any blockers in not being able to reuse? 

Remember you can always do something like 

`<section class=""namespace-sectionname> (or) <div>` and the target the ampstart- classes inside them to a different style.

You will have to lookout for semantics & accessibility (you don't have to worry about this if you reuse)

css/templates/gallery/page.css
```diff
@@ -21,9 +21,85 @@
...
+}
+
+.gallery-base-invert {
+  color: #fff;
+  mix-blend-mode: difference;
+  -webkit-font-smoothing: antialiased;
+}
+
+/* Hack media query to target IE, which does not support mix-blend-mode. */
+@media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
```
camelburrito: move this to the bottom, also i am not sure if :none is a value that is valid

https://msdn.microsoft.com/en-us/library/hh771830(v=vs.85).aspx

css/templates/gallery/components/_landing-images.css
```diff
@@ -0,0 +1,581 @@
...
+  background-image: url("data:image/svg+xml;charset=utf-8,%3Csvg xmlns='http://www.w3.org/2000/svg' width='18' height='18' fill='%23000' viewBox='0 0 18 18'%3E%3Cpath d='M9 3L7.94 4.06l4.19 4.19H3v1.5h9.13l-4.19 4.19L9 15l6-6z'/%3E%3C/svg%3E");
+  right: 0;
+}
+
+.amp-carousel-button-prev {
+  left: 0;
+  background-image: url("data:image/svg+xml;charset=utf-8,%3Csvg xmlns='http://www.w3.org/2000/svg' width='18' height='18' fill='%23000' viewBox='0 0 18 18'%3E%3Cpath d='M15 8.25H5.87l4.19-4.19L9 3 3 9l6 6 1.06-1.06-4.19-4.19H15v-1.5z'/%3E%3C/svg%3E");
+}
+
+@media (max-width: 40.06rem) {
```
camelburrito: This should already be defined , use the existing variables

css/templates/gallery/components/_landing-images.css
```diff
@@ -0,0 +1,581 @@
...
+  .amp-carousel-button {
```
camelburrito: This is too generic of a name for the carousel - it should be namespaced like in the instructions
something like `gallery-landing-wall-*`

Just because this is the only carousel on this page does not mean users will not add more carousels, 

If you want to make all the carousel buttons have the same look and feel this approach is valid, how ever i doubt that is what is going on here.

css/templates/gallery/components/_landing-images.css
```diff
@@ -0,0 +1,581 @@
...
+  min-width: 100vw;
+  overflow: hidden;
+}
+
+.gallery-landing-wall-scroll::after {
+  width: 0px;
+  height: 100%;
+}
+
+@media (max-width: 40.06rem) {
```
camelburrito: Use the definitions here
https://github.com/ampproject/ampstart/blob/master/css/ampstart-base/bass-override-vars.css

css/templates/gallery/components/_landing-images.css
```diff
@@ -0,0 +1,581 @@
...
+  .gallery-landing-wall-scroll {
+    margin-bottom: -3px; /* remove spacing from display inline-block */
+  }
+}
+
+@keyframes hover-one {
```
camelburrito: give better names to keyframes (namespaced + meaningful)

css/templates/gallery/components/_landing-images.css
```diff
@@ -0,0 +1,581 @@
...
+  100% {transform: translateY(-2vh) rotateZ(360deg); }
+}
+
+@keyframes hover-two {
+  0% {transform: translateY(-1vh) rotateZ(360deg); }
+  50% {transform: translateY(1vh) rotateZ(360deg); }
+  100% {transform: translateY(-1vh) rotateZ(360deg); }
+}
+
+[class^="gallery-landing-image-wrap-"] {
```
camelburrito: `[class^="gallery-landing-image-wrap-"]`
I would generally avoid going this route. AMP does not recommend any one to write styles like this (even though it is not said out loud)
I think it is easier to add another class

Please fix elsewhere too

css/templates/gallery/components/_landing-images.css
```diff
@@ -0,0 +1,581 @@
...
+}
+
+[class^="gallery-landing-image-wrap-"] amp-img {
+  object-fit: cover;
+  box-shadow: 1vh 6vh 5vh var(--color-shadow);
+  transition: transform 300ms;
+  pointer-events: none;
+}
+
+.gallery-landing-image-wrap-0 amp-img { background: #d97b7f }
```
camelburrito: overall it is easier to use a generic class name and :nth-child

css/templates/gallery/page.css
```diff
@@ -21,9 +21,85 @@
...
+@import './components/_landing-images';
 
 body {
   color: var(--font-color-primary);
   font-family: var(--font-family-primary);
   line-height: var(--line-height);
 }
+
+.gallery-base-content {
+  background: #fff;
```
camelburrito: i think some basic colors are defined as variables (white and black), prefer to use the variables.

css/templates/gallery/components/_landing-images.css
```diff
@@ -0,0 +1,581 @@
...
+
+@media (--breakpoint-sm) {
+  .gallery-landing-title {
+    font-size: 7vw;
+    padding: 1rem 5vw;
+  }
+}
+
+.amp-carousel-button {
+  background-color: white;
```
camelburrito: hex (prefer to reuse --color-white)

css/templates/gallery/components/_landing-images.css
bentranter: On the final version of this template, there's an "inner page" (see here https://ampstart-staging-metalab.firebaseapp.com/templates/gallery/gallery-inner.amp) that uses the same carousel -- would you rather we use the current approach, or should I modify this carousel along with the one on the inner page as you suggested?

css/templates/gallery/page.css
bentranter: Neal might be able to help here @10xjs

templates/gallery/partials/landing-images.snip.html
bentranter: I'm not sure -- Neal? (cc @10xjs?)

templates/gallery/partials/landing-images.snip.html
bentranter: This was done deliberately, so that the target for clicking an image is slightly smaller than the image itself. This was necessary because some of the images overlap.

templates/gallery/partials/landing-images.snip.html
bentranter: You're right that it is really similar, but we really need the `{{#positions}}` classes injected into the `<figcaption>` element, and that isn't doable with the "image-with-caption" component like it is with our custom code.

templates/gallery/partials/landing-images.snip.html
camelburrito: I see , thanks for the explanation!

css/templates/gallery/components/_landing-images.css
```diff
@@ -0,0 +1,581 @@
...
+.gallery-landing-image-wrap:nth-child(16) amp-img { background: #aeb197 }
+.gallery-landing-image-wrap:nth-child(17) amp-img { background: #370afa }
+.gallery-landing-image-wrap:last-child amp-img { background: #c1c1bf }
+
+.gallery-landing-image-wrapped:hover amp-img {
+  transform: translateY(-1vh) scale(1.05);
+  transition: transform 200ms ease-in-out;
+}
+
+.gallery-landing-image-wrap:first-child {
```
camelburrito: can you move the background colors inside the bigger definitions, that should save some bytes.

templates/gallery/partials/landing-images.snip.html
camelburrito: if that is the only reason , then it is easy to solve this

0. Use nth child - IMO having numbered classes is really a bad idea.
1. Pass in the full class via JSON (add a fig caption class inside the image component)

10xjs: @camelburrito Beyond the fact that the markup contains similar tags (not even in the same structure) this shares little in common with the `image-with-caption` component.

Switching to `image-with-caption` would mean all of the styles would have to be re-written to adapt to the new markup structure. The current structure is directly coupled to the interactive animations and image transformations. There is a risk that the markup structure of `image-with-caption` entirely incompatible with the desired effects.

css/templates/gallery/components/_landing-images.css
bentranter: The background colours are targeting the `amp-img` within the `.gallery-landing-image-wrap` though, moving them to the bigger definitions would break functionality.

templates/gallery/partials/landing-images.snip.html
camelburrito: I see, that makes some sense , lets proceed with this then

PR #682, opened 2017-10-20
[GALLERY] Add inner page and map

bentranter:
Adds the inner page for the Gallery template, which also loads the map via an iframe. Staging link: https://metalab-gallery-inner-page-cdcb8897c18.surge.sh/templates/gallery/gallery-inner.amp.html.

cc @camelburrito


templates/gallery/gallery-inner.amp.html
```diff
@@ -0,0 +1,207 @@
...
+            8,359.5 Z
+          "
+          stroke="none"
+          fill="#F5F6FD"
+        ></path>
+      </svg>
+      {{> ./partials/hero.snip.html}}
+      <section class="gallery-base-section gallery-base-rich-text">
+        <div class="gallery-base-container mx-auto">
+          <div class="clearfix">
```
camelburrito: you will not need to use clearfix if you use flexboxes - which is super easy in bass, any reason why we are using floats (.col)

templates/gallery/gallery-inner.amp.html
```diff
@@ -0,0 +1,207 @@
...
+                  alt="pink lamp on a blue wall"
+                ></amp-img>
+              </div>
+            </div>
+          </div>
+        </div>
+      </section>
+      <section class="gallery-base-section gallery-base-rich-text">
+        <div class="gallery-base-container mx-auto">
+          <div class="clearfix">
```
camelburrito: ditto

templates/gallery/gallery-inner.amp.html
```diff
@@ -0,0 +1,207 @@
...
+                "
+                alt="mountain at sunset"
+              ></amp-img>
+            </div>
+          </div>
+        </div>
+      </section>
+      <section class="gallery-base-section gallery-base-rich-text">
+        <div class="gallery-base-container mx-auto">
+          <div class="clearfix">
```
camelburrito: ditto

templates/gallery/gallery-inner.amp.html
```diff
@@ -0,0 +1,207 @@
...
+                alt="blue lake against skyline"
+              ></amp-img>
+            </div>
+          </div>
+        </div>
+      </section>
+      <section class="gallery-base-section gallery-base-rich-text">
+        <div class="gallery-base-container mx-auto">
+          <h2 class="gallery-base-invert relative z1 pb2 mb2">Other Exibits</h2>
+          <ul class="related-section-items list-reset flex flex-wrap m0">
```
camelburrito: namespace the class

templates/gallery/gallery-inner.amp.html
```diff
@@ -0,0 +1,207 @@
...
+            <li class="col-12 sm-col-4 md-col-4 lg-col-4 pr2 py2">
+              <figure class="image-with-caption m0 relative mb4">
```
camelburrito: use the ampstart-image-with-cation here for sure

templates/gallery/gallery-inner.amp.html
camelburrito: I believe you did not want the entire `ampstart-related-section` is that why you are not using it here?

templates/gallery/gallery-inner.amp.html
```diff
@@ -0,0 +1,207 @@
...
+                <amp-img src="../../img/gallery/plant.jpg" width="233" height="202" layout="responsive" class="gallery-recommended-image" alt="house plant on a chair on a yellow background"></amp-img>
+                <figcaption class="h5 mt1 px3">
+                  Untitled &mdash; Evan Ondre
+                </figcaption>
+              </figure>
+            </li>
+            <li class="col-12 sm-col-4 md-col-4 lg-col-4 pr2 py2">
```
camelburrito: i see repeating HTML , can you move the data into the JSON and create a loop

templates/gallery/gallery-map.html
```diff
@@ -0,0 +1,318 @@
...
+          icon: {
+            url: 'data:image/svg+xml;charset=UTF-8;base64,' + btoa(document.getElementById('marker').innerHTML),
+            size: new google.maps.Size(80, 80),
+            origin: new google.maps.Point(0, 0),
+            anchor: new google.maps.Point(40, 40)
+          }
+        });
+      }
+    </script>
+    <!-- NOTE: Maps API key is from https://ampbyexample.com/components/amp-iframe/ -->
```
camelburrito: Open a bug to create a maps API for ampstart
Add a todo for me here link to the bug

templates/gallery/gallery-map.html
```diff
@@ -0,0 +1,318 @@
+<!doctype html>
```
camelburrito: create a folder called 3p inside this template and move this file in there (The only stuff that needs to be there inside gallery is packable AMP HTML)

templates/gallery/partials/map.snip.html
```diff
@@ -0,0 +1,43 @@
...
+      </div>
+    </div>
+    <div class="gallery-component-map-frame">
+      <amp-iframe
+        height="300"
+        layout="fixed-height"
+        sandbox="allow-scripts allow-top-navigation"
+        scrolling="no"
+        frameborder="0"
+        src="gallery-map.html?marker={{marker.lat}},{{marker.lng}}"
```
camelburrito: We might run into same origin issues if we do this in preview mode, prefix this with https://ampstart-iframes.firebaseapp.com/ (and the full path in dist)

templates/gallery/partials/hero.snip.html
```diff
@@ -0,0 +1,75 @@
...
+limitations under the License.
+-->
+}}
+
+{{#hero}}
+  <header class="gallery-component-hero clearfix">
+    <div class="gallery-base-container mx-auto">
+      <div class="gallery-component-hero-content absolute left-0 right-0 z1">
+        <h1
+          class="gallery-component-hero-heading"
```
camelburrito: formatting is all weird in this file , please fix

templates/gallery/partials/button.snip.html
```diff
@@ -0,0 +1,33 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS-IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+}}
+
+{{#button}}
+  <{{#href}}a{{/href}}{{^href}}div{{/href}}
```
camelburrito: why div and not `<button>`?

templates/gallery/partials/hero.snip.html
```diff
@@ -0,0 +1,75 @@
...
+            width="{{width}}"
+            height="{{height}}"
+          ></amp-img>
+        {{/images}}
+      </amp-carousel>
+    </div>
+    {{/carousel}}
+    {{#content}}
+    <div class="gallery-base-container gallery-base-rich-text mx-auto">
+      <div class="clearfix">
```
camelburrito: use flexboxes to get rid of this clearfix

templates/gallery/gallery-inner.amp.json
```diff
@@ -0,0 +1,77 @@
...
+          "src": "../../img/gallery/hall.jpg",
+          "2x": "../../img/gallery/hall@2x.jpg",
+          "alt": "colourful hallway"
+        }]
+      },
+      "content": "<blockquote>Justo libero ut nunc consequat massa. Vitae et nulla. Mus arcu in. Tellus at ut. Ac mi varius ut.</blockquote>"
+    },
+    "map": {
+      "heading": "Visit the exhibition from <time datetime=\"2017-08-24\">Sept 24</time>-<time datetime=\"2017-08-29\">29</time>",
+      "subheading": "123 St Name <br/>City, State 12345",
```
camelburrito: This is not subheading ,this belongs in an address tag, and please try not to use BRs

templates/gallery/gallery-map.html
bentranter: Do you want the bug opened against this repo or that main amp repo?

templates/gallery/gallery-inner.amp.html
bentranter: Refactoring these to use flex layout is a significant amount of work.

All the responsive column will need to be recreated from scratch, since BassCSS doesn't have a flex equivalent of the `col col-<n>` system. This will also lead to a lot more custom CSS, since we'd need to essentially recreate the responsive grid system from scratch using flexbox.

If implemented we'd have classes like the one below all over the codebase.

```css
@media (--breakpoint-md) {
  .gallery-flex-md-col-9 {
    flex: 1 1 75%;
  }
}
```

templates/gallery/gallery-map.html
bentranter: Do you mean to just move the file to `templates/gallery/3p/gallery-map.html`, and other than, continue using as-is? Or do I need to make a modification to this file?

templates/gallery/partials/map.snip.html
bentranter: I'll do this, but I'm not sure how to do this so that it works on `localhost` dev environments as well as production. Is there any reference code I can look at?

templates/gallery/gallery-inner.amp.html
camelburrito: That is not correct - 

http://basscss.com/#basscss-flexbox
There is plenty of instances where col-* is used with flexboxes

infact you take a look a ampstart.com or some of our components we follow exactly this approach.

templates/gallery/gallery-inner.amp.html
camelburrito: response?

templates/gallery/gallery-map.html
camelburrito: Yes

templates/gallery/gallery-map.html
camelburrito: This Repo

templates/gallery/partials/map.snip.html
camelburrito: It wont work on localhost. It is a problem that ill have to fix :). For now we will need to make it work on staging and production. I can push to that domain once this PR is in and localhost would look good

css/templates/gallery/components/_button.css
```diff
@@ -0,0 +1,44 @@
...
+  font-size: 1.1rem;
+  letter-spacing: 0.02em;
+  font-weight: 600;
+  border: 1px solid var(--color-primary);
+  color: var(--color-white);
+  transition: transform 0.1s, box-shadow 0.1s;
+  box-shadow: 0 0.3rem 1rem rgba(6, 7, 22, 0.1);
+}
+
+button.ampstart-btn {
```
camelburrito: do  you need this specificity, also are you sure you want to set it to 0?

css/templates/gallery/components/_hero.css
```diff
@@ -0,0 +1,122 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+.gallery-component-hero {
+  margin-bottom: calc(60 / 16rem);
```
camelburrito: all calculations need to be documented well or use variables to demonstrate how we arrived at this value. Otherwise this will cause UI to break even with small changes

css/templates/gallery/page.css
```diff
@@ -80,6 +83,215 @@ body {
   font-weight: 600;
 }
 
+/**
```
camelburrito: copyright should be on top of the file?

templates/gallery/gallery-map.html
```diff
@@ -0,0 +1,318 @@
...
+
+<!---
```
camelburrito: copyright should be the first thing on the file

templates/gallery/gallery-inner.amp.html
bentranter: Ah that's amazing! I had been looking at the v7 docs, my mistake -- I'll get this fixed ASAP.

css/templates/gallery/components/_button.css
bentranter: Yes, since there's a conditional in our template when we either set that class on an `<a>` element, or a `<button>` element. The `button` element needs the line-height removed to be vertically centred, but applying that to the `a` element breaks it, so I've scoped the class.

templates/gallery/gallery-map.html
bentranter: I'll do this, and while it shouldn't break content detection for most clients (https://mimesniff.spec.whatwg.org/ lists `<!--` as an indication that the document's content type is HTML), it's possible it'll break some.

PR #683, opened 2017-10-20
[GALLERY] Add header to landing page

bentranter:
Adds the header component to the landing page. Staging link: https://metalab-gallery-landing-header-87bf.surge.sh/templates/gallery/gallery.amp

cc @camelburrito

### DIFF
--- a/templates/gallery/partials/header.snip.html
+++ b/templates/gallery/partials/header.snip.html
@@ -18,15 +18,6 @@ limitations under the License.
 
 {{#header}}
   <header class="gallery-header {{#invert}}gallery-base-invert{{/invert}} {{#mobile-header}}gallery-mobile-header{{/mobile-header}}">
-    <button class="ampstart-btn gallery-header-menu-button" role="button" tabindex="-1" on="tap:sidebar.open">
-      <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">
-        <g fill="currentColor" fill-rule="evenodd">
-          <rect width="98" height="6" x="1" y="19"/>
-          <rect width="98" height="6" x="1" y="47"/>
-          <rect width="98" height="6" x="1" y="75"/>
-        </g>
-      </svg>
-    </button>
     <div class="gallery-header-rotate">
       <div class="gallery-header-heading">
         {{#href}}<a href="{{href}}">{{/href}}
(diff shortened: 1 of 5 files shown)


templates/gallery/partials/header.snip.html
```diff
@@ -0,0 +1,49 @@
...
+}}
+
+{{#header}}
+  <header
+    class="
+      gallery-header
+      {{#invert}}gallery-base-invert{{/invert}}
+      {{#mobile-header}}gallery-mobile-header{{/mobile-header}}
+    "
+  >
```
camelburrito: why is this formatted weird - the closing arrow is in a new line? don't line break on HTML it will render weird in the view source and it could also cause issues that are hard to debug.

templates/gallery/partials/header.snip.html
```diff
@@ -0,0 +1,49 @@
...
+    <button
+      class="ampstart-btn gallery-header-menu-button"
```
camelburrito: are you going to be using your snipper from previous PR here?

templates/gallery/partials/header.snip.html
```diff
@@ -0,0 +1,49 @@
...
+      role="button"
+      tabindex="-1"
+      on="tap:sidebar.open"
+    >
+      <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">
```
camelburrito: may be move to img file and import from there - Like we did here https://github.com/ampproject/ampstart/blob/master/components/social-follow/social-follow.snip.html#L33

templates/gallery/partials/header.snip.html
bentranter: No, this is the button to open the sidebar, I actually think this will be removed in the sidebar PR once we refactor the component you suggested -- I'll get back to you on this once I look at the sidebar PR again.

templates/gallery/partials/header.snip.html
bentranter: Similar to the above comment, I think will change because of our Sidebar PR changes, I'll update here once I figure out the plan.

bentranter:
I need to figure out the Sidebar situation before this can be merged, since we might be removing the `<button>` from this PR.

templates/gallery/partials/header.snip.html
camelburrito: I guess if comments are addressed we can merge this and you can always delete this on future PRs. This makes sense if there are other useful parts in this PR

PR #751, opened 2017-11-05
Fix Beck & Galo image paths on menu page

robinpokorny:
When downloaded, Beck & Galo menu page has no images.

It is because the paths to images are wrong by one level.


camelburrito:
@robinpokorny - Thanks for the fix !! Appreciate your contribution!

Issue #767, opened 2017-11-15
[Gallery] Loaders - remove 3 dot loaders


spacedino:
Turn off the 3 dot loader can be over written with "no loading" attribute. https://www.ampproject.org/docs/reference/common_attributes#noloading

The custom loaders you have are great without the 3 dots.

ask @camelburrito if you have an issue with this one - he things its an easy and quick fix.

davidtapp:
@spacedino @camelburrito same goes here as per my note on #766.

camelburrito:
Adding the ‘noloading’ attribute should get rid of the 3 dots

Issue #766, opened 2017-11-15
[Travel] Loader - turn off 3 dot loader


spacedino:
Turn off the 3 dot loader can be over written with "no loading" attribute. https://www.ampproject.org/docs/reference/common_attributes#noloading

The custom loaders you have are great without the 3 dots.

ask @camelburrito if you have an issue with this one - he things its an easy and quick fix.

davidtapp:
@spacedino @camelburrito Hey team, we're just trying to be hyper focused on the ship blocking bugs that have sprung up post-PRs. We'll put this one in the enhancement bucket we'll probably leave for now based on a need to come to a close on this asap.

Issue #724, opened 2017-11-02
[Gallery - Exhibit Page] Update location on map at page bottom


mattludwig:
Should we change the map at the page bottom to have a red location dot somewhere other than the Google HQ? @spacedino @ericlindley-g

mattludwig:
FYI - I don't believe I have access to assign or put this issue in the Metalab project.

ericlindley-g:
+1   and /cc @camelburrito on the GitHub access issue

davidtapp:
Will do this one. Changing it to the Met! 1000 5th Ave, New York, NY 10028, USA

Issue #714, opened 2017-11-01
[Gallery - Sidebar] Missing "Gallery of Modern Art"


spacedino:
**Current**
![image](https://user-images.githubusercontent.com/7339798/32255708-78df0d18-be67-11e7-822b-bea78daa56ab.png)

**Should be**
![image](https://user-images.githubusercontent.com/7339798/32255713-80420614-be67-11e7-911b-13886df4105b.png)

spacedino:
This fix broke it even further. 

**Should be**
"Gallery of Modern Art" should be vertical, to match the Home page. It was like this before....don't know why this changed. 

https://user-images.githubusercontent.com/7339798/32255713-80420614-be67-11e7-911b-13886df4105b.png


**Current**
"Gallery of Modern Art" is across the top of the page horizontally.
![image](https://user-images.githubusercontent.com/7339798/32817226-798dbede-c971-11e7-8d9b-1c4a2c91b4fb.png)
![image](https://user-images.githubusercontent.com/7339798/32817235-844e1dd2-c971-11e7-992d-daa8b8e30825.png)

PR #775, opened 2017-11-21
[GALLERY + TRAVEL] Fix remaining issues

bentranter:
Provides fixes for:

* #766
* #767
* #724
* #714
* #713

cc @camelburrito

### DIFF
--- a/css/templates/gallery/page.css
+++ b/css/templates/gallery/page.css
@@ -291,16 +291,6 @@ body {
   .v-lg-hide { display: none }
 }
 
-.gallery-hide-mobile {
-  display: none;
-}
-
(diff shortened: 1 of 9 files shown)


css/templates/gallery/page.css
```diff
@@ -291,6 +291,16 @@ body {
   .v-lg-hide { display: none }
 }
 
+.gallery-hide-mobile {
```
camelburrito: can't you use sm-hide and/or xs-hide

bentranter: Oh yeah, I forgot those were part of `core.css`. Done!

Issue #785, opened 2017-12-08
[TRAVEL] - Launch blocking bugs FINAL


spacedino:
Home page - match mock up

- [x] Replace tiny "Discover more" text with "Explore"

- [x] Center "Explore" text"

- [x] Move carat below "Explore" text - 8px padding

- [x] Make "Explore" + carat tappable > should trigger page scroll to top

- [x] Fix dead white area > should be darker to match mock

![image](https://user-images.githubusercontent.com/7339798/33788505-8fecba9a-dc27-11e7-9e4a-2a1cfcdc5ad8.png)
 >>>>>>>>>>
![image](https://user-images.githubusercontent.com/7339798/33788514-9dbfd6de-dc27-11e7-8404-3a498005d9c7.png)

- [x] Move "Discover adventures" heading to the bottom, next to the blog section (this matches the desktop info grouping model. Placement will remain at the bottom of the page, this is different than the destop version)

- [x] Add sub heading below "discover adventures" to match desktop (this is missing)
![image](https://user-images.githubusercontent.com/7339798/33788600-17a61b16-dc28-11e7-8208-f7ff0c3a607c.png)

- [x] Add card behind blog section. see mock.

- [x] Make sure image + blog section are vertically centered. Right now they are not.

![image](https://user-images.githubusercontent.com/7339798/33788640-3c4dc4dc-dc28-11e7-9e94-0ba1f92a80f2.png)

>>>>>>>>>

![image](https://user-images.githubusercontent.com/7339798/33788653-45e24c66-dc28-11e7-99f0-fee9bafef83f.png)

Results page

- [x] remove 3 dot loader that is appearing over purple header

Mock

![17 copy 2x](https://user-images.githubusercontent.com/7339798/33788671-586b1dc2-dc28-11e7-9a42-557ab1767045.png)

spacedino:
LGTM - @ericlindley-g @mattludwig to approve.

Issue #784, opened 2017-12-07
Correct rel=canonical link for product-listing.amp


ericlindley-g:
Currently:

https://ampstart.com/templates/e-commerce/product-listing.amp
... has a rel=canonical set to:
https://www.ampstart.com/templates/e-commerce/product-details.amp

Should be
https://www.ampstart.com/templates/e-commerce/product-listing.amp

PR #790, opened 2017-12-12
Template fixes

camelburrito:
Closes https://github.com/ampproject/ampstart/issues/784

Partially Fixes https://github.com/ampproject/ampstart/issues/785

### DIFF
--- a/templates/travel/partials/hero/hero.snip.html
+++ b/templates/travel/partials/hero/hero.snip.html
@@ -142,8 +142,8 @@ limitations under the License.
           Find Adventures & Tours
         </a>
       <!--/ Search Form -->
-      <a class="travel-hero-discover block center mx-auto mt1 md-hide lg-hide">
-        Discover adventures {{> ../../../../img/travel/icon/caret-down-small.svg}}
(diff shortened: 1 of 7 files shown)


templates/travel/partials/hero/hero.snip.html
```diff
@@ -142,8 +142,8 @@ <h2 class="travel-hero-subheading line-height-2 bold xs-hide sm-hide">Find and b
           Find Adventures & Tours
         </a>
       <!--/ Search Form -->
-      <a class="travel-hero-discover block center mx-auto mt1 md-hide lg-hide">
-        Discover adventures {{> ../../../../img/travel/icon/caret-down-small.svg}}
+      <a class="travel-hero-discover block center mx-auto mt1 md-hide lg-hide" on="tap:travel-landing-content.scrollTo">
```
aghassemi: still keep an `href="#travel-landing-content"` in addition to the scrollTo to support "Open in new window" and copy link cases.

camelburrito: Thanks ! Good catch!

PR #824, opened 2018-08-13
Added <amp-consent> stuff for GDPR

morsssss:
I wasn't sure whether the `components` dir was the best place to stash the consent HTML and CSS. For example, www/partials contains other partials like `footer.snip.html`.  But I liked the idea of putting the HTML and CSS in the same dir.  Let me know if you want this moved elsewhere.

### DIFF
--- a/components/consent/consent.css
+++ b/components/consent/consent.css
@@ -1,0 +1,101 @@
+@keyframes slideUp {
+  0% {
+    transform: translateY(100%);
+    opacity: 0.5;
+  }
+  100% {
+    transform: translateY(0);
+    opacity: 1;
+  }
+}
+
+amp-consent {
+  background: white;
+  box-shadow: rgba(0, 0, 0, 0.26) 0 0 19px 0;
+  animation: .5s ease-out 0s 1 slideUp;
+}
+
+.consent-popup a {
+  text-transform: none;
+  font-size: inherit;
+  text-decoration: none;
+}
+
+.consent-popup a.close-button {
+  position: absolute;
+  color: #333;
+}
+
+.consent-popup .ampstart-btn {
+  width: 150px;
+  text-align: center;
+}
+
+.consent-popup p {
+  line-height: 20px;
+  font-size: 14px;
+}
+
+.consent-popup p a {
+  color: #b60845;
+}
+
+.consent-popup div {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.consent-popup button {
+  padding: 13px 18px;
+  width: 142px;
+  flex: none;
+  align-self: center;
+}
+
+/* phones */
+@media (max-width: 640px) {
+  amp-consent a.close-button {
+    font-size: 22px;
+    top: 15px;
+    right: 13px;
+  }
+  
+  .consent-popup div {
+    flex-flow: column wrap;
+    align-items: center;
+    margin: 12px 25px 15px 12px;
+  }
+
+  .consent-popup p {
+    padding: 5px 20px 5px 10px;
+    margin: 0 0 10px 0;
+  }
+}
+
+/* tablets and desktop */
+@media (min-width: 641px) {
+ .consent-popup a.close-button {
+    font-size: 25px;
+    left: 30px;
+  }
+
+  .consent-popup div {
+    flex-flow: row;
+    height: 100px;
+    margin: 0 40px;
+  }
+
+  .consent-popup p {
+    max-width: 660px;
+    padding: 0 25px 0 40px;
+    margin: 0;
+  }
+}
+
+@media (max-width: 320px) {
+  .consent-popup p {
+    line-height: 16px;
+    font-size: 12px;
+  }
+}
--- a/components/consent/consent.snip.html
+++ b/components/consent/consent.snip.html
@@ -1,0 +1,32 @@
+<amp-geo layout="nodisplay">
+  <script type="application/json">
+    {
+      "ISOCountryGroups": {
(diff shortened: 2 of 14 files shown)


morsssss:
@camelburrito , are you still working on this site?  If so, can you review?  If not, who should I ask?

camelburrito:
ill review today

components/consent/consent.css
```diff
@@ -0,0 +1,101 @@
...
+    opacity: 0.5;
+  }
+  100% {
+    transform: translateY(0);
+    opacity: 1;
+  }
+}
+
+amp-consent {
+  background: white;
```
camelburrito: all colors should be in terms of variables (theme parameters)

components/consent/consent.css
```diff
@@ -0,0 +1,101 @@
...
+  box-shadow: rgba(0, 0, 0, 0.26) 0 0 19px 0;
+  animation: .5s ease-out 0s 1 slideUp;
+}
+
+.consent-popup a {
+  text-transform: none;
```
camelburrito: all of these should be done in the markup , using the bass css utility

components/consent/consent.css
```diff
@@ -0,0 +1,101 @@
...
+  line-height: 20px;
+  font-size: 14px;
+}
+
+.consent-popup p a {
+  color: #b60845;
+}
+
+.consent-popup div {
+  display: flex;
```
camelburrito: all alignments etc , should be in the markup (for the baseline theme)

components/consent/consent.css
```diff
@@ -0,0 +1,101 @@
...
+
+.consent-popup button {
+  padding: 13px 18px;
+  width: 142px;
+  flex: none;
+  align-self: center;
+}
+
+/* phones */
+@media (max-width: 640px) {
```
camelburrito: media queries should be from predefined constants

components/consent/consent.css
morsssss: Sorry. I should have known that!  Will re-work.

components/consent/consent.css
```diff
@@ -0,0 +1,71 @@
...
+@media (--breakpoint-sm) {
+ .ampstart-consent-popup a.ampstart-consent-close-btn {
+    font-size: 25px;
+    left: 30px;
+  }
+
+  .ampstart-consent-popup div {
+    flex-flow: row;
+    height: 100px;
+    margin: 0 40px;
```
camelburrito: use spacing vars for margins (unless they are one off)

components/consent/consent.css
```diff
@@ -0,0 +1,71 @@
...
+}
+
+/* tablets and desktop */
+@media (--breakpoint-sm) {
+ .ampstart-consent-popup a.ampstart-consent-close-btn {
+    font-size: 25px;
+    left: 30px;
+  }
+
+  .ampstart-consent-popup div {
```
camelburrito: there does not seem to be any semantic in this selector, since you are targeting a div its prone to break fast. recommend you to add a classname if that is easier

components/consent/consent.css
```diff
@@ -0,0 +1,71 @@
...
+    flex-flow: row;
+    height: 100px;
+    margin: 0 40px;
+  }
+
+  .ampstart-consent-popup p {
+    max-width: 41.25em;
+    padding: 0 25px 0 40px;
```
camelburrito: use spacing variables

components/consent/consent.css
```diff
@@ -0,0 +1,71 @@
...
+    margin: 0;
+  }
+}
+
+@media (max-width: 320px) {
+  .ampstart-consent-popup p {
+    line-height: 1em;
+    font-size: .75em;
```
camelburrito: both this should be taken care by the typography classes?

components/consent/consent.css
morsssss: Ok! I was afraid to add any more classnames, but I can do that, following the style already used on the site.

Thanks for the lightning-fast response, by the way ⚡️⚡️⚡️

components/consent/consent.css
morsssss: I wasn't sure how to do that within a media query...

components/consent/consent.css
morsssss: Ok!  I've added vars for spacing, line height, and font size where it seemed to me very close to the existing design and where I could apply them consistently.  Let me know if you want moooooreeeeeee...

camelburrito: in general all margins and paddings should use the variables (unless they  are exemptions by design)

components/consent/consent.css
camelburrito: yeah there are no responsive typography classes - may be discuss with abby on this  (non blocking) but use the --h* variables (look in the typography.css) instead of the actual values. 

FYI: we dont use ems - we use rems.

morsssss:
Ok... how's this?

I left px values in two places, but everywhere else I've used CSS vars where I had them, and rem units where I didn't.

components/consent/consent.css
```diff
@@ -34,30 +34,30 @@ amp-consent .ampstart-btn {
...
     flex-flow: column wrap;
     align-items: center;
-    margin: 12px 25px 15px 12px;
+    margin: .75rem var(--space-2) var(--space-1) .75rem;
   }
 
   .ampstart-consent-popup p {
-    padding: 5px 20px 5px 10px;
-    margin: 0 0 10px 0;
+    padding: .25rem 1.25rem .25rem .625rem;
```
camelburrito: https://github.com/ampproject/ampstart/blob/master/css/ampstart-base/bass-override-vars.css#L25

You should use one of these especially for margins/padding/outlines 

since the component could get used in different templates and each template will define what their spacing theme is, you should use the spacing vars.

morsssss:
thank you!

lmk if you want me to do the push to the appropriate environments.

PR #842, opened 2018-11-19
fix npm vulnerabilities

camelburrito:


torch2424:
Tested manually, and everything appears to be working 😄

Issue #862, opened 2019-01-10
Ability to specify the amp component with specific version


marcelovani:
When I run the output on the AMP Validator, it shows an error saying:

> The extension 'amp-mustache' is referenced at version '0.1' which is a deprecated version. Please use a more recent version of this extension. This may become an error in the future.

See: https://www.ampproject.org/docs/reference/components/amp-mustache

I want to suggest the ability to specify the version of the amp components

jpettitt:
See https://github.com/ampproject/amphtml/blob/master/spec/amp-versioning-policy.md

marcelovani:
So what I am suggesting is that we can specify the version for each component, see https://github.com/ampproject/ampstart/pull/863/files

jpettitt:
Ah, that makes sense.

PR #863, opened 2019-01-10
#862 Provide component and version

marcelovani:
This is just a suggestion to solver the issue with different versions of components.
This is not the full task.
If it's decided that we should do this, we can go on and update the other json files.

### DIFF
--- a/utils/amp-extension-script.snip.html
+++ b/utils/amp-extension-script.snip.html
@@ -17,9 +17,9 @@ limitations under the License.
 }}
 
 {{#extensions}}
-<script custom-element="{{.}}" src="https://cdn.ampproject.org/v0/{{.}}-0.1.js" async></script>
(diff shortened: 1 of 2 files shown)


utils/amp-extension-script.snip.html
```diff
@@ -17,9 +17,9 @@
 }}
 
 {{#extensions}}
-<script custom-element="{{.}}" src="https://cdn.ampproject.org/v0/{{.}}-0.1.js" async></script>
+<script custom-element="{{component}}" src="https://cdn.ampproject.org/v0/{{component}}-{{version}}.js" async></script>
```
camelburrito:
```suggestion
<script custom-element="{{component}}" src="https://cdn.ampproject.org/v0/{{component}}-{{#version}}{{.}}{{/version}}{{^version}}0.1{{/version}}.js" async></script>
```

sebastianbenz:
Closing in favor of AMP Start 2.0

PR #866, opened 2019-03-03
AMP START

Bhanditz:
Gdpr
Open


alankent:
Hi @sebastianbenz! Does ampstart.com still exist? If not, do you think we can close this PR? Sounds like GDPR related for when ampstart.com was built from this repo. (Now the ampstart.com URLs redirect to amp.dev I believe.)

sebastianbenz:
ampstart redirect now, so this is no longer needed

PR #870, opened 2019-04-08
add referrer

sebastianbenz:


alankent:
Hi @sebastianbenz, since ampstart.com seems to be working out (doing redirects), do you think we still need this PR accepted? I can review and accept if still useful.

Issue #877, opened 2019-06-09
ReferenceError: primordials is not defined


amejiarosario:
Using node 12 and gulp 3. That combination does not work: https://github.com/gulpjs/gulp/issues/2324

```
$ npm run build

> kickstart@ build /Users/admejiar/Code/tmp/ampstart
> gulp build

fs.js:27
const { Math, Object } = primordials;
                         ^

ReferenceError: primordials is not defined
    at fs.js:27:26
```

I tried gulp 4 and getting different errors

amphyploid:
Regarding the Node version: the [`engines` section of the package.json](https://github.com/ampproject/ampstart/blob/master/package.json#L4) specifies that Node version `^7.0.0` is required. The caret (`^`) allows changes that do not modify the left-most non-zero digit in the [major, minor, patch] tuple. Therefor, the highest version one should use is [v7.10.1](https://nodejs.org/en/blog/release/v7.10.1/). You'll need to use [NVM](https://github.com/nvm-sh/nvm) or [NVM for Windows](https://github.com/coreybutler/nvm-windows) to switch versions and build as intended.

Regarding the Gulp version: the [`devDependencies` section of the package.json](https://github.com/ampproject/ampstart/blob/master/package.json#L46) specifies that Gulp version `^3.9.1` is required. One should use the locally-installed version of Gulp rather than the global installation. Having a different version of Gulp installed globally may interfere with using the locally installed one. The global installation should probably be uninstalled.

From what I saw from other bug reports, "ReferenceError: primordials is not defined" seems to primarily affect Windows users.  Even with the correct versions of the above, additional bugs are present on Windows 10. @amejiarosario, if you are on Windows 10, you will need to wait for #878 to be merged before `npm i` and `npm run build` will complete smoothly. Feel free to clone and use my fork in the mean time.

PR #878, opened 2019-06-24
Make cross-platform portable

amphyploid:
This pull request addresses #877 in the "Lock `first-mate` subdependency to v7.3.0" commit and additional path-related errors on Windows in the "Update path segment separators to be cross-platform" commit. Details are provided in their respective commit descriptions. It was tested on Windows 10 and Debian 9.


tasks/bundle.js
```diff
@@ -64,12 +64,13 @@ function collectResources(filepath, html, templateName, done) {
...
     imgs.forEach(function(imgpath) {
       if (imgpath) {
-        let imgFile = imgpath.split('img/')[1];
+        let imgFile = imgpath.split('img' + path.sep)[1];
         const dest = `.archive/${templateName}/img/${imgFile}`;
         fs.copySync(imgpath, dest);
       }
     });
-    let pathToTmpl = filepath.replace(/.*templates\/(.*)/, '\$1');
+    const pathToTmplRegex = new RegExp('.*templates\\' + path.sep + '(.*)');
```
Phrohdoh: I don't believe this is correct.

In the previous implementation, with regex literal, the `\/` is escaping the `/` so it isn't seen as the end of the regex.

The intent isn't to have a literal `\/` or `\\` in the string, which this change will do cause.

amphyploid: It's quite confusing, thanks for allowing me the opportunity to explain. In this situation, `\\` is actually `\` escaping `\` so that it doesn't cause the `'` to be escaped. This `\` is then used to escape the path segment separator that `path.sep` returns as both (`\` and `/`) are characters that need to be escaped in JS strings. It took me a while to remember why I did this. 😉

4mfie:
@googlebot I signed it!

package.json
```diff
@@ -42,6 +42,7 @@
     "eslint-plugin-babel": "^4.0.1",
     "eslint-plugin-google-camelcase": "0.0.2",
     "extract-text-webpack-plugin": "^2.0.0-rc.3",
+    "first-mate": "7.3.0",
```
sebastianbenz: Why is this dependency needed?

4mfie: That's explained in the [commit message](https://github.com/4mfie/ampstart/commit/703419abf380000f431bf023c709d3254b989c03). I'll paste its contents below for the sake of convenience.

The `highlights` package depends on a version range of the `first-mate` package, which currently resolves to v7.4.0. This version of `first-mate` depends on `oniguruma` v7.2.0. This version of `oniguruma` is broken. I am locking `first-mate` at v7.3.0, which uses the previous functioning version of `oniguruma` (v7.1.0) in order for `npm i` to complete.

alankent:
Hi @sebastianbenz - just a ping on whether you want to accept or reject this PR so we can close it out. Thanks!

sebastianbenz:
Let's close this for now, as we're going to deprecate the old template build process anyway. The should make sure though that the new version works cross platform.

PR #892, opened 2019-12-20
Auto extension import

sebastianbenz:
Two related changes:

1. Support extension import via attribute

Some extensions, such as amp-video-docking, add new attributes (e.g.
`dock`) to existing AMP tags (e.g. amp-video). This commit will
automatically import the required extensions for these attributes.

This add support for:

* amp-video-docking
* amp-fx-collection (only for AMP tags)
* amp-lightbox-gallery

2. amp-bind auto-import

* Adds support for auto-importing the amp-bind extension based on the
presence of a binding.
* Adds a new binding syntax for AMP tags: `bindAttribute` instead of
`data-amp-bind-attribute`.

Caveat: both only work on AMP tags, but not on non-AMP tags such as
`div`.


lib/next-amp-demo/pages/components/AmpBind.tsx
```diff
@@ -0,0 +1,37 @@
...
+export default () => (
+  <>
+    <h1>amp-bind</h1>
+    <AmpCarousel bindSlide='slide' width='400' height='300' type='slides'>
+      <AmpImg src='https://picsum.photos/id/231/400/300' layout='fill'></AmpImg>
+      <AmpImg src='https://picsum.photos/id/232/400/300' layout='fill'></AmpImg>
+      <AmpImg src='https://picsum.photos/id/233/400/300' layout='fill'></AmpImg>
+    </AmpCarousel>
+    <button
+      on='tap:AMP.setState({
```
alankent, line 30: Do we care about indentation of multi-line attributes? Prettier won't touch them unfortunately.

lib/next-amp-demo/pages/components/AmpLightboxGallery.tsx
```diff
@@ -0,0 +1,91 @@
...
+  <main>
+    <h1>amp-image-lightbox</h1>
+    <h2>With AmpImg</h2>
+    <AmpImg
+      lightbox=''
+      src='https://picsum.photos/id/237/300/200'
+      width='300'
+      height='200'
+      layout='responsive'
+    ></AmpImg>
```
alankent, line 32: Do we have a style guide for whether to use `<AmpImg... />` vs `<AmgImg...></AmpImg>`?

lib/next-amp-demo/pages/components/AmpBind.tsx
sebastianbenz: I personally care a lot about these. Would be great if we could find some kind of standard here.

lib/next-amp-demo/pages/components/AmpLightboxGallery.tsx
sebastianbenz: I'd prefer `<AmpImg... />, I'm only writing the other out of habit...

PR #895, opened 2020-01-10
Support bindings on non-AMP elements

sebastianbenz:
Going with a wrapper based approach for none AMP elements for now:

```
<AmpBind bindText="greeting">
  <div />
</AmpBind>
```

I don't see a different way to support the `bindX` shortcut for non-AMP
elements.

### DIFF
--- a/lib/next-amp/README.md
+++ b/lib/next-amp/README.md
@@ -215,14 +215,28 @@ Better than Option 1, but makes things more complicated.
 
 ### amp-bind
 
-For `amp-bind` expressions, the `amp-bind` extension is automatically imported using the same approach as for `amp-fx-collection`. The following will work out-of-the-box:
+AMP bindings can be expressed using the the `bindX` short-form instead of `data-amp-bind-x`. A nice
+side-effect is that most JS IDEs will offer component specific code-completion once you've
+typed `bind`.
+
+Here is an example binding an `AmpImg`'s `src` attribute via `bindSrc`:
 
 ```
-<div data-amp-bind-text="hello"/>
-<button on="tap:AMP.setState({ "hello": "world" })">
+<AmpImg width="300" height="200" src="image.png" bindSrc="imageUrl"/>
+<button on="tap:AMP.setState({ imageUrl: 'anotherImage.png' })">Change image</button>
 ```
 
-State can be initialized via:
+Unfortunately, this doesn't work for non-AMP tags such as `div` or `input`. In this case you need to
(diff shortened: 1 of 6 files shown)


lib/next-amp/src/components/AmpBind.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Renders an amp-state element, by either adding local state via `value`
+ * or remote state via the `src` property.
+ */
+const AmpBind: React.FunctionComponent<AmpBindProps> = ({children, ...rest}) => {
+  const newProps = {};
+  Object.keys(rest).forEach(key => {
+    if (/^bind[A-Z]/g.test(key)) {
+      const newKey = `data-amp-bind-${key.substring(4).toLowerCase()}`;
```
alankent, line 29: Do you want to lower case the whole string or just the first character of the string? I know standard HTML elements won't have hyphenated names, but just felt a little risking assuming nothing else ever will.

lib/next-amp/README.md
```diff
@@ -215,14 +215,28 @@ Better than Option 1, but makes things more complicated.
...
+<button on="tap:AMP.setState({ imageUrl: 'anotherImage.png' })">Change image</button>
+```
+
+Unfortunately, this doesn't work for non-AMP tags such as `div` or `input`. In this case you need to
+wrap the element inside the `AmpBind` component:
 
 ```
-<div data-amp-bind-text="hello"/>
-<button on="tap:AMP.setState({ "hello": "world" })">
+<AmpBind bindText="greeting">
```
alankent, line 233: It is a pity extra markup is required find bind attributes, but if there is no alternative supported by Next.js then a wrapper element names sense.

lib/next-amp/src/components/AmpBind.tsx
sebastianbenz: The React toString serializer will automatically lowercase attribute names. So I think it's fine to lowercase by default.

lib/next-amp/README.md
sebastianbenz: I'm not happy about it either. I see two other options:

1. HTML Element specific components:
    ```<AmpDiv bindText/>```
   This would have the added benefit of element specific auto completion. But it's probably more confusing than it helps. 
2. Add `bindX` types to the HTML elements and rename the attributes in AMP Optimizer. This would work best, but the downside is that it means we have to move Next.js specific features into AMP Optimizer.

PR #896, opened 2020-01-15
Restructured to be closer to HTML partials version.

alankent:
This PR is not intended to be merged. 

This still has lots of data-amp-bind-* in it, without loading up AmpBind. But it shows progress of renaming all the components to align with the HTML partials version.

It also moves all the AMP State variables from global state to a single nested group. This is not the final version - it should have multiple global namespaces (not one per variable, not one for everything).

@sebastianbenz


templates/travel/components/activities/Activities.js
```diff
@@ -0,0 +1,61 @@
...
+          Browse
+          <br />
+          by activity
+        </h3>
+        <h3 className='bold h1 line-height-2 mb2 xs-hide sm-hide center'>Browse by activity</h3>
+      </div>
+      <div className='overflow-scroll'>
+        <div className='travel-overflow-container'>
+          <div className='flex justify-center p1 md-px1 mxn1'>
+            {[
```
sebastianbenz, line 26: suggestion: extract the list into a variable

sebastianbenz: Seems also to duplicate TravelData

templates/travel/components/data/TravelData.js
```diff
@@ -0,0 +1,202 @@
+export default function TravelData() {
```
sebastianbenz, line 1: This feels very hard to read and it also seems to mix different kinds of data (domain and presentational data). I think it'd be better to split this up into multiple files.

templates/travel/components/data/TravelData.js
```diff
@@ -0,0 +1,202 @@
...
+          {text: 'Nature', value: 'nature', icon: {pine: true}},
+          {text: 'Nightlife', value: 'nightlife', icon: {moon: true}},
+          {text: 'Tours', value: 'tours', icon: {bus: true}},
+          {text: 'Water', value: 'water', icon: {boat: true}},
+        ],
+      },
+      'head-tag': {
+        title: 'Travel Template',
+        'canonical-path': 'templates/travel/travel.amp',
+        extensions: ['amp-list', 'amp-bind'],
```
sebastianbenz, line 21: this is no longer needed

templates/travel/components/data/TravelData.js
```diff
@@ -0,0 +1,202 @@
...
+          {text: 'Music', value: 'music', icon: {guitar: true}},
+          {text: 'Nature', value: 'nature', icon: {pine: true}},
+          {text: 'Nightlife', value: 'nightlife', icon: {moon: true}},
+          {text: 'Tours', value: 'tours', icon: {bus: true}},
+          {text: 'Water', value: 'water', icon: {boat: true}},
+        ],
+      },
+      'head-tag': {
+        title: 'Travel Template',
+        'canonical-path': 'templates/travel/travel.amp',
```
sebastianbenz, line 20: this isn't needed either

templates/travel/components/data/TravelResultsData.js
```diff
@@ -0,0 +1,146 @@
+export default function TravelResultsData() {
```
sebastianbenz, line 1: general question: do we want to shorten default exports to `export default () => { ... }`?

templates/travel/components/utils/AmpListProps.js
```diff
@@ -0,0 +1,16 @@
+export default function AmpListProps(includeDates) {
+  return {
+    src:
+      '/api/search?maxPrice=0' +
```
sebastianbenz, line 4: Use `URL` and `URL.searchParams` to create the string instead. It's much more readable.

templates/travel/components/utils/AmpListProps.js
```diff
@@ -0,0 +1,16 @@
+export default function AmpListProps(includeDates) {
```
sebastianbenz, line 1: Maybe find a better name for this function / file. It's not a util.

templates/travel/pages/api/foo.js
```diff
@@ -0,0 +1,378 @@
...
+
+/**
+ * cors is HTTP middleware that sets an AMP-specific cross origin resource
+ * sharing HTTP header on our response.
+ *
+ * @param {Object} req - The HTTP request object.
+ * @param {Object} res - The HTTP response object.
+ * @param {Function} next - The next HTTP handler to execute.
+ */
+function cors(req, res, next) {
```
sebastianbenz: use @ampproject/toolbox-cors instead.

templates/travel/pages/index.js
```diff
@@ -36,492 +42,81 @@ export default function Index(props) {
...
-        <AmpState id='fields_sort_edited'>{false}</AmpState>
-        <AmpState id='query_sort'>{'popularity-desc'}</AmpState>
-
-        <AmpState id='fields_maxPrice'>{801}</AmpState>
-        <AmpState id='fields_maxPrice_initial'>{801}</AmpState>
-        <AmpState id='fields_maxPrice_live'>{801}</AmpState>
-        <AmpState id='fields_maxPrice_edited'>{false}</AmpState>
-        <AmpState id='query_maxPrice'>{801}</AmpState>
-      </div>
+      <AmpState id='display'>
```
sebastianbenz, line 50: I'm not a fan of namespacing via name as it results in unnecessary redundancy. Instead I'd prefer:

```
<AmpState id="ui">{{
   hero: true,
   reset: false,
  ...
}}<AmpState/>
```

templates/travel/pages/results.js
```diff
@@ -17,1137 +17,103 @@ limitations under the License.
...
+        />
+        <meta name="amp-google-client-id-api" content="googleanalytics" />
+        <link
+          href="https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,600,700"
+          rel="stylesheet"
+        />
+        <link rel="stylesheet" type="text/css" href="/css/travel-results.css" />
+      </Head>
+
+      <AmpState id="display">
```
sebastianbenz, line 47: this feels redundant

sebastianbenz (commented):
I think it's already much better. Left a few comments.

templates/travel/components/activities/Activities.js
alankent: Some I converted from result of converter and then reverse engineered the source template. Next one I think should port the Mustache files rather than the output of the conversion. There are few like this around the place I have not done yet.

templates/travel/components/data/TravelData.js
alankent: This was copied out of the original template - I have not got to the bottom of the two sets of data - feels like some overlap there as well. Let's work out what we want as components to be listed separately - that might also help clean all this up as well.

